PR #107, opened 2026-04-23
fix(bolt): honor database scoped rollback

### LINKED ISSUE #106: [BUG] Database-scoped Bolt explicit transactions break rollback semantics
## Bug Description
Database-scoped Bolt explicit transactions can leak autocommit semantics when a Neo4j driver uses managed writes (`ExecuteWrite`) against a named database.

In the failing path, an earlier write inside the managed transaction can remain visible after a later statement fails and the driver rolls the transaction back. The same area also affects recursive execution paths that re-enter `executeInTransaction`, because re-wrapping an already active transaction can drop namespace metadata for multi-database write shapes.

## Steps to Reproduce
1. Build and run NornicDB with Bolt enabled.
2. Connect with the Neo4j Go driver to a named database such as `nornic`.
3. Execute a managed write transaction that first creates a node and then runs an invalid statement.
4. Query for the created node after the transaction returns an error.

Minimal reproduction using the Neo4j Go driver:

```go
ctx := context.Background()
driver, err := neo4j.NewDriverWithContext("bolt://127.0.0.1:7687", neo4j.NoAuth())
if err != nil {
    panic(err)
}
defer driver.Close(ctx)

session := driver.NewSession(ctx, neo4j.SessionConfig{
    AccessMode:   neo4j.AccessModeWrite,
    DatabaseName: "nornic",
})
defer session.Close(ctx)

_, err = session.ExecuteWrite(ctx, func(tx neo4j.ManagedTransaction) (any, error) {
    result, runErr := tx.Run(ctx, "CREATE (n:TxProbe {id: $id})", map[string]any{"id": "driver-rollback"})
    if runErr != nil {
        return nil, runErr
    }
    if _, consumeErr := result.Consume(ctx); consumeErr != nil {
        return nil, consumeErr
    }

    result, runErr = tx.Run(ctx, "THIS IS NOT CYPHER", nil)
    if runErr != nil {
        return nil, runErr
    }
    _, consumeErr := result.Consume(ctx)
    return nil, consumeErr
})
fmt.Printf("managed write error: %v\n", err)

verify, err := session.Run(ctx, "MATCH (n:TxProbe {id: $id}) RETURN count(n) AS c", map[string]any{"id": "driver-rollback"})
if err != nil {
    panic(err)
}
record, err := verify.Single(ctx)
if err != nil {
    panic(err)
}
fmt.Printf("count after rollback: %v\n", record.Values[0])
```

4. For the namespace variant, run a multi-step `UNWIND ... MATCH ... MERGE` write shape inside an explicit transaction against a namespaced database and verify generated IDs stay correctly prefixed.

## Expected Behavior
- The invalid statement should cause the managed transaction to fail.
- The earlier `CREATE` should be rolled back completely.
- Re-entrant execution inside the same explicit transaction should reuse the active transaction wrapper and preserve namespace metadata.

## Actual Behavior
- The managed transaction returns an error, but the earlier write can survive and remain visible after rollback.
- Recursive execution paths can re-wrap the same active transaction and lose namespace context for generated IDs in multi-database write flows.

## Environment
- **NornicDB Version**: `v1.0.42-hotfix` (also reproduced on the pre-fix fork state before the merged rollback patch)
- **OS**: macOS (Apple Silicon)
- **Architecture**: ARM64
- **Docker Image**: built from source
- **Go Version** (if built from source): `go1.26.2`

## Cypher Query (if applicable)
```cypher
CREATE (n:TxProbe {id: $id})
THIS IS NOT CYPHER
MATCH (n:TxProbe {id: $id}) RETURN count(n) AS c
```

## Error Messages / Logs
```text
[BOLT] RUN user=anonymous remote=127.0.0.1:<port> db=nornic status=ERROR rows=0 ... err=syntax error: query must start with a valid clause (MATCH, CREATE, MERGE, DELETE, CALL, SHOW, EXPLAIN, PROFILE, ALTER, USE, BEGIN, COMMIT, ROLLBACK, etc.)
```

The problem is not the syntax error itself. The problem is that the earlier successful write can survive even though the managed transaction fails and the client expects rollback semantics.

## Additional Context
A fork-local fix and regression coverage were just merged here:
- `linuxdynasty/NornicDB` PR `#1` (`fix(bolt): honor database scoped rollback`)

That patch does two things:
- routes database-scoped Bolt `BEGIN` through a true transactional executor instead of cached/autocommit execution
- reuses the existing transaction storage wrapper during recursive execution so namespace metadata is preserved

Regression coverage added in the fork:
- clean rollback in database-scoped Bolt executor
- rollback after statement failure
- Neo4j-driver `ExecuteWrite` rollback
- direct `executeInTransaction` wrapper-reuse regression for the `transaction.go` path

I can open an upstream PR from the fork with the same narrow fix and tests if this issue direction looks right.

## Checklist
- [x] I have searched existing issues to ensure this is not a duplicate
- [x] I have tested this with the latest version of NornicDB
- [x] I have included all relevant information (environment, query, error messages)

linuxdynasty:
## Summary
Fix database-scoped Bolt explicit transactions so Neo4j-driver managed transactions use real Cypher `BEGIN`, `COMMIT`, and `ROLLBACK` semantics instead of cached autocommit execution.

Closes #106.

## Why This Change Was Needed
The database-scoped Bolt path could fall back into cached/autocommit-style execution, which meant:
- rollback-sensitive managed writes could leak partial state after a later statement failed,
- recursive `UNWIND ... MATCH ... MERGE` execution could lose namespace metadata if the same active transaction was wrapped twice,
- and the regression harness could hide startup failures or race on `server.listener` while waiting for the test server to bind.

## Changes
- route Bolt `BEGIN` through a non-cached transactional database-scoped executor
- reuse the active transaction storage wrapper during recursive execution so namespace metadata is preserved across re-entrant query paths
- harden the Bolt integration harness so startup failures fail loudly and test port discovery no longer races on `server.listener`
- add direct regression coverage for the transaction-wrapper reuse path, alongside rollback and canonical `UNWIND/MATCH/MERGE` write coverage

## Why The New Tests Matter
- `pkg/bolt/server_dbmanager_integration_test.go` now proves the external Neo4j-driver path sees correct rollback behavior without relying on racy listener polling.
- `pkg/cypher/transaction_wrapper_reuse_test.go` proves the internal `transaction.go` recursion path reuses the existing wrapper from context, which is the invariant that keeps multi-database namespace prefixes intact during fallback execution.

## Testing
- `go test -tags 'noui nolocalllm' ./pkg/bolt -run 'TestSessionGetExecutorForDatabase_Rollback|TestDatabaseManagerBoltNeo4jDriverExecuteWrite' -count=1 -v`
- `go test -tags 'noui nolocalllm' ./pkg/cypher -run 'TestExecuteInTransaction_ReusesExistingTransactionWrapper' -count=1 -v`
- `go test -tags 'noui nolocalllm' ./pkg/bolt ./pkg/cypher ./pkg/txsession -count=1`
- `git diff --check`

### DIFF
--- a/pkg/bolt/server.go
+++ b/pkg/bolt/server.go
@@ -1906,7 +1906,7 @@ func (s *Session) handleBegin(data []byte) error {
 		if dbName == "" {
 			dbName = s.server.dbManager.DefaultDatabaseName()
 		}
-		txExec, err := s.getExecutorForDatabase(dbName)
+		txExec, err := s.getTransactionalExecutorForDatabase(dbName)
 		if err != nil {
 			return s.sendFailure("Neo.ClientError.Database.DatabaseNotFound",
 				fmt.Sprintf("Database '%s' not found: %v", dbName, err))
@@ -2372,10 +2372,6 @@ func (s *Session) getExecutorForDatabase(dbName string) (QueryExecutor, error) {
 		return nil, fmt.Errorf("database manager not available")
 	}
 
-	type authAwareStorageResolver interface {
-		GetStorageWithAuth(name string, authToken string) (storage.Engine, error)
-	}
-
 	useAuthScopedResolver := false
 	if resolver, ok := s.server.dbManager.(authAwareStorageResolver); ok && strings.TrimSpace(s.forwardedAuthHeader) != "" {
 		_ = resolver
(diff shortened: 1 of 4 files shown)


linuxdynasty:
On a side note, I love what you are doing with this project. I'm currently trying to migrate off of Neo4j to NornicDB

orneryd:
Thanks dude! looks good! assuming the CI passes i'll pull it in!

orneryd (approved):

PR #115, opened 2026-04-25
Use unique index for constraint validation

linuxdynasty:
## Summary
- use rebuilt single-property UNIQUE constraint value indexes during Badger transaction validation
- keep the existing label-scan fallback when the unique-value cache is not authoritative
- add a regression hook/test proving hot validation does not fall back to a label scan

## Why
Large schema-constrained MERGE batches were spending most CPU in full label scans during UNIQUE validation. Once schema reload has rebuilt authoritative unique values, validation can use the registered value index instead.

## Tests
- go test -tags noui,nolocalllm ./pkg/storage -run 'TestBadgerTransactionUniqueConstraintUsesRegisteredValueIndex|TestBadgerTransaction_FullScanUniqueConstraint' -count=1 -v

### DIFF
--- a/pkg/cypher/schema.go
+++ b/pkg/cypher/schema.go
@@ -471,7 +471,11 @@ func (e *StorageExecutor) executeCreateConstraint(ctx context.Context, cypher st
 		if err := storage.ValidateConstraintOnCreationForEngine(e.storage, constraint); err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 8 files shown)


orneryd:
Thanks for this! the API surface is really large so I really apreciate the help 🫶

linuxdynasty:
You got my support!

orneryd:
hop in discord sometime @linuxdynasty

PR #116, opened 2026-04-25
Route UNWIND MATCH map merges through hot path

linuxdynasty:
## Summary
- route supported UNWIND MATCH/OPTIONAL MATCH mutation tails through the generalized merge-chain hot path only when the plan proves supported
- add SET n += row.props map-merge handling inside the chain executor
- preserve existing fixed-chain and generic MATCH routing for unsupported shapes
- document the lookup-then-upsert map-merge write shape in the hot-path cookbook

## Why
Bounded ingestion pipelines often resolve a parent with MATCH, upsert a child with MERGE, merge a row property map, and create an idempotent relationship. Without this route, those statements can fall back to per-row execution or miss the specialized hot path.

## Tests
- go test -tags noui,nolocalllm ./pkg/cypher -run TestUnwindMergeBatch_MatchMergeSetMapPropertiesUsesHotPath -count=1 -v
- go test -tags noui,nolocalllm ./pkg/cypher -run 'TestUnwindMergeBatch|TestE2E_UnwindMergeChainBatch|TestUnwindBenchHopLinkBatch' -count=1
- go test -tags noui,nolocalllm ./pkg/cypher -run 'TestUnwindMergeBatch_MatchMergeSetMapPropertiesUsesHotPath|TestBatchRelationshipMergeWithUnwindKeys|TestMigrationShape_UnwindMatchMergeSetMap_ComplexRowsAndClauses|TestBug_UnwindMatchCreate_CorrelatedJoinKeepsCardinality|TestUnwindTopLevelMergeSetMergeUsesNestedMapPropertyAccess' -count=1
- git diff --check

### DIFF
--- a/docs/performance/hot-path-query-cookbook.md
+++ b/docs/performance/hot-path-query-cookbook.md
@@ -559,7 +559,24 @@ MERGE (event)-[:BELONGS_TO]->(key)
 
 Use this for qVersions-style write families where one input row materializes multiple node upserts and multiple idempotent relationship upserts in a deterministic order.
 
-### 7.3e.1 Staged Compound `UNWIND` Version Pipeline
+### 7.3e.1 Lookup-Then-Upsert With Map-Merge Properties
+
+```cypher
+UNWIND $rows AS row
+MATCH (parent:EntityA {primaryKey: row.parentKey})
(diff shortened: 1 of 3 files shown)


orneryd:
thank you for updating the cookbook too!

PR #118, opened 2026-04-25
Use unique constraints for MERGE lookup

linuxdynasty:
## Summary
- add a schema lookup API for single-property unique constraint values
- route MERGE node lookup through unique constraint metadata before label-scan fallback
- add regression coverage that forbids label/full scans when a unique constraint can resolve the node

## Why
PCG dogfood runs create uid uniqueness constraints such as Function.uid and then issue batched MERGE writes on that uid. NornicDB was validating uniqueness through unique constraint metadata, but findMergeNode only consulted composite/property indexes before falling back to label scans. As a graph grows, this makes each MERGE slower even when a unique constraint already proves the key is indexed.

## Validation
- go test -tags noui,nolocalllm ./pkg/cypher -run 'TestFindMergeNode_UsesUniqueConstraintLookupWithoutScan|TestFindMergeNode_UsesPropertyIndexLookup|TestExecuteMerge_RecoversFromDuplicateCreateForParameterizedRepositoryMerge' -count=1 -v
- git diff --check

### DIFF
--- a/pkg/cypher/merge.go
+++ b/pkg/cypher/merge.go
@@ -210,6 +210,25 @@ func (e *StorageExecutor) findMergeNode(store storage.Engine, labels []string, p
 			}
 		}
 
+		for _, prop := range mergePropertyNamesSorted(props) {
+			val := props[prop]
+			nodeID, valueFound, constraintExists := schema.LookupUniqueConstraintValue(label, prop, val)
+			if !constraintExists {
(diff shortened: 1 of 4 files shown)


orneryd:
this one has conflicts now, once you resolve them I'll pull it in 🫶

orneryd (approved):

PR #119, opened 2026-04-25
Optimize batched relationship writes for lookup-first UNWIND chains

linuxdynasty:
## Summary

This adds a batched mutation path for lookup-first UNWIND chains shaped like:

```cypher
UNWIND $rows AS row
MATCH (source:CodeEntity|Function {id: row.source_id})
MATCH (target:CodeEntity|Table {id: row.target_id})
MERGE (source)-[rel:QUERIES]->(target)
SET rel.repo_id = row.repo_id
```

PCG emits this form for SQL relationship materialization: rows already identify existing source/target nodes, and the write only needs to merge/update the relationship. Before this patch, NornicDB fell back to the generic per-row UNWIND MATCH path because the batch detector did not accept lookup-first MATCH mutation chains, label alternatives, relationship variables, or relationship property SET assignments.

## Why

On a PCG dogfood run against the platform-context-graph repository, SQL relationship materialization took about `217.718543222s` for only 291 SQL edges. After this patch, the same phase on the 16-vCPU test host completed in `3.921462s`. That moved the remaining bottleneck back to semantic entity materialization instead of relationship writes.

## Safety

- Keeps pure read shapes such as `UNWIND ... MATCH ... RETURN` out of the mutation batch path.
- Restricts the new route to chains that include a relationship MERGE mutation.
- Supports the existing batch execution machinery rather than adding a separate executor.

## Validation

```bash
go test -tags 'noui nolocalllm' ./pkg/cypher -run 'TestUnwindMatch(RelationshipSetBatch_PCGSQLRelationshipShape|ReadDoesNotUseMutationBatchPath)' -count=1
go test -tags 'noui nolocalllm' ./pkg/cypher ./pkg/storage ./pkg/bolt -count=1
go build -tags 'noui nolocalllm' -o ./bin/nornicdb-headless-pcg-sql-edge-hotpath ./cmd/nornicdb
```

Runtime validation: remote PCG dogfood on the 16-vCPU test host with this binary showed SQL relationship materialization dropping from `217.718543222s` to `3.921462s`, with owner startup healthy at about `3m06s`.

### DIFF
--- a/pkg/cypher/cache.go
+++ b/pkg/cypher/cache.go
@@ -935,9 +935,10 @@ func makeLookupKey(label string, props map[string]interface{}) string {
 //   - The cached node, or nil if not in cache
 func (e *StorageExecutor) lookupCachedNode(label string, props map[string]interface{}) *storage.Node {
 	key := makeLookupKey(label, props)
-	e.nodeLookupCacheMu.RLock()
+	cacheMu := e.nodeLookupCacheLock()
+	cacheMu.RLock()
 	node := e.nodeLookupCache[key]
-	e.nodeLookupCacheMu.RUnlock()
+	cacheMu.RUnlock()
 	return node
 }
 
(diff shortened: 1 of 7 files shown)


orneryd:
Keep em coming :D but seriously we should coordinate next time. I am going to try and regularly drop new versions AT LEAST maintenance versions as dependencies come through dependabot, so people should be able to rely on the release cycle and maintenance updates. Anything that comes in before the next dependabot (or i get a wild hair on a new feature 😁)

linuxdynasty:
lol, well, since I am really trying to get NornicDB working with my project, you should see maybe a few more PR's coming in as I migrate from Neo4j to using NornicDB.

linuxdynasty:
Addressed the Copilot review comments in 5121f55:

- any-label MERGE lookups now fall back to label scans when schema lookups miss an unindexed alternative label, and the redundant AllNodes fallback was removed
- relationship MERGE batch notifications now fire only when the edge is created or updated
- CreateEdge ErrAlreadyExists now refetches a concurrent relationship when present, otherwise retries with a new edge ID
- added regression coverage for unindexed alternative labels, no-op MERGE notifications, ID collision retry, and concurrent relationship create/update

orneryd:
i'll get to this tomorrow probably as i'm out and about today.

orneryd:
pushing to latest docker tag

PR #120, opened 2026-04-26
Map Bolt transaction conflicts as transient

linuxdynasty:
## Summary

This PR makes NornicDB's Neo4j-compatible write path behave like Neo4j for
retryable transactional conflicts while keeping the batch mutation hot paths
safe under concurrent PCG graph writes.

It does three related things:

- Classifies enumerated MVCC write conflicts, resource-pressure snapshot
  failures, and deadlock-style transaction failures as Neo4j transient
  transaction errors so official Neo4j drivers can retry them instead of
  treating them as terminal client/syntax failures.
- Shares node lookup cache locking across transaction clones so concurrent
  transactional execution does not race on cached merge lookups.
- Preserves named relationship variables and `SET` assignments in
  `UNWIND ... MERGE relationship` batch execution, which PCG uses for canonical
  relationship writes.

The shared transient-error logic now lives in `pkg/errors` so both the Bolt
server path and HTTP/server execution path can use the same classification
without depending on a Neo4j-specific package name. Classification is based on
error identity via `errors.Is` and shared sentinels, not localized message text.

## Why this matters

PCG writes canonical graph data through Neo4j-compatible drivers. Under
concurrent reducer/projector load, optimistic MVCC conflicts are expected
contention signals, not query syntax errors. Neo4j drivers already know how to retry
`Neo.TransientError.Transaction.*` failures; NornicDB needs to surface those
conditions with the same semantics so callers can make forward progress without
application-specific error string matching.

Without this, concurrent graph writes can fail terminally even though the safe
behavior is to retry the transaction. That makes NornicDB look less compatible
than Neo4j for real ingestion workloads and forces downstream applications to
paper over a database-level classification issue.

## Changes

- Adds a shared transient transaction error classifier in `pkg/errors`.
- Exposes shared error sentinels/aliases for transaction conflicts, MVCC
  pressure, snapshot expiration, and transaction deadlock classification.
- Maps Bolt implicit transaction conflicts and deadlock-style failures to
  retryable Neo4j transient transaction codes.
- Maps HTTP transaction endpoint failures through the same sentinel-based
  classifier.
- Keeps explicit `COMMIT` fallback behavior for non-transient commit failures.
- Synchronizes cloned transaction lookup caches to avoid data races during
  concurrent merge execution.
- Supports named relationship variables with relationship `SET` assignments in
  `UNWIND MERGE` batch execution.
- Hardens Bolt failure test decoding so tests assert the protocol-visible error
  shape reliably.

## Verification

- `go test -tags 'noui nolocalllm' ./pkg/errors ./pkg/bolt ./pkg/server ./pkg/cypher -count=1`
- Focused Bolt conflict/commit tests covering transient mapping and explicit
  commit fallback behavior.
- Focused HTTP transaction error-mapping tests covering sentinel-based
  classification without message substring matching.
- Focused Cypher batch tests covering named relationship assignment behavior.

### DIFF
--- a/pkg/bolt/server_extra_test.go
+++ b/pkg/bolt/server_extra_test.go
@@ -12,6 +12,8 @@ import (
 	"time"
 )
 
+// failureCodeFromResponse decodes a chunked Bolt FAILURE response and returns
+// its Neo4j-compatible error code.
 func failureCodeFromResponse(t *testing.T, data []byte) string {
 	t.Helper()
 	if len(data) < 4 {
@@ -48,6 +50,8 @@ func failureCodeFromResponse(t *testing.T, data []byte) string {
 	return code
 }
 
+// TestFailureCodeFromResponseConcatenatesChunks verifies that test assertions
+// inspect the complete Bolt payload instead of only the first response chunk.
 func TestFailureCodeFromResponseConcatenatesChunks(t *testing.T) {
 	payload := append([]byte{0xB1, MsgFailure}, encodePackStreamMap(map[string]any{
 		"code":    "Neo.TransientError.Transaction.DeadlockDetected",
--- a/pkg/bolt/server_extra_test.go
+++ b/pkg/bolt/server_extra_test.go
@@ -17,11 +17,23 @@ func failureCodeFromResponse(t *testing.T, data []byte) string {
 	if len(data) < 4 {
 		t.Fatalf("response too short: %d bytes", len(data))
 	}
-	chunkSize := int(data[0])<<8 | int(data[1])
-	if len(data) < 2+chunkSize {
-		t.Fatalf("response chunk truncated: size=%d bytes=%d", chunkSize, len(data))
+
+	var payload []byte
+	for offset := 0; ; {
+		if len(data[offset:]) < 2 {
+			t.Fatalf("response missing chunk header at offset %d: bytes=%d", offset, len(data))
+		}
+		chunkSize := int(data[offset])<<8 | int(data[offset+1])
+		offset += 2
+		if chunkSize == 0 {
+			break
+		}
+		if len(data[offset:]) < chunkSize {
+			t.Fatalf("response chunk truncated: size=%d offset=%d bytes=%d", chunkSize, offset, len(data))
+		}
+		payload = append(payload, data[offset:offset+chunkSize]...)
+		offset += chunkSize
 	}
-	payload := data[2 : 2+chunkSize]
 	if len(payload) < 2 || payload[0] != 0xB1 || payload[1] != MsgFailure {
 		t.Fatalf("response is not a FAILURE message: %x", payload)
 	}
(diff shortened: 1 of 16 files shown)


pkg/bolt/server_extra_test.go
```diff
@@ -12,6 +12,30 @@ import (
...
+func failureCodeFromResponse(t *testing.T, data []byte) string {
+	t.Helper()
+	if len(data) < 4 {
+		t.Fatalf("response too short: %d bytes", len(data))
+	}
+	chunkSize := int(data[0])<<8 | int(data[1])
+	if len(data) < 2+chunkSize {
+		t.Fatalf("response chunk truncated: size=%d bytes=%d", chunkSize, len(data))
+	}
+	payload := data[2 : 2+chunkSize]
```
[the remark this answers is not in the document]
linuxdynasty: Fixed in 5847364. failureCodeFromResponse now concatenates all Bolt chunks until the zero-size terminator before decoding FAILURE metadata, and TestFailureCodeFromResponseConcatenatesChunks covers a split response.

orneryd:
I think for this instead of having a "neo4j compat" package, i'd prefer it if it were abstracted to more something like a generic errors package that I can start moving all of the existing errors into in order to manage them more effectively rolling forward, plus when we move all the error messages into a single location we can then easily localize it into various languages. What do you think?

edit: I appreciate the concept of neo4j compat, we are by definition meant to be neo4j compatible by default so I'd remove that compat concept as a layer itself.

orneryd (changes requested):
comments in the thread.

linuxdynasty:
> I think for this instead of having a "neo4j compat" package, i'd prefer it if it were abstracted to more something like a generic errors package that I can start moving all of the existing errors into in order to manage them more effectively rolling forward, plus when we move all the error messages into a single location we can then easily localize it into various languages. What do you think?
> 
> edit: I appreciate the concept of neo4j compat, we are by definition meant to be neo4j compatible by default so I'd remove that compat concept as a layer itself.

I definitely agree with you on this. The neo4jcompat package is gone now and the shared transient transaction mapping now lives in a generic pkg/errors package, with package docs framed around centralized public error codes/classification and future localization. Bolt and HTTP now import it as nornicerrors, and I added package-level tests for the mapping.

pkg/errors/errors.go
```diff
@@ -0,0 +1,32 @@
...
+// MapTransientTransactionError maps conflict/deadlock failures to retryable transaction codes.
+func MapTransientTransactionError(message string) (string, bool) {
+	m := strings.ToLower(strings.TrimSpace(message))
+	if m == "" {
+		return "", false
+	}
+	if strings.Contains(m, "deadlock") {
+		return TransientDeadlockDetected, true
+	}
+	if strings.Contains(m, "changed after transaction start") ||
```
orneryd: for these guys I don't think we should map them based on the text strings for production code. they should be enumarated inside the errors pachage and mapped directly via reference rather than the message content which will be localized. If the errors are strings or templates we should move them here as part of this refactor, what do you think?

orneryd (changes requested):

pkg/errors/errors.go
linuxdynasty: I agree with you as well. My bad for not thinking about that.

linuxdynasty:
I updated this in 9c67cce. I agreed with your feedback @orneryd , so the classifier no longer looks at error message text. It now takes the actual error and uses errors.Is against the shared sentinels/aliases in pkg/errors.

The protocol response still uses err.Error() for the message, so the user-facing text stays the same. Only the retry classification logic changed.

linuxdynasty:
Rebased this on the latest main after the relationship-probe PR landed. The conflict was in the overlapping UNWIND MERGE hot-path code, so I kept the merged batch-local relationship cache and the named relationship SET support together.

I also reran the affected suite after the rebase:

```bash
go test -tags 'noui nolocalllm' ./pkg/errors ./pkg/bolt ./pkg/server ./pkg/cypher -count=1
```

GitHub now reports the branch as no longer dirty; it is just waiting on review/check status.

orneryd (approved):

PR #124, opened 2026-04-27
Optimize UNWIND merge relationship probes

linuxdynasty:
## Summary

This PR optimizes the `UnwindMergeChainBatch` hot path for ingestion workloads
that create a target node in the current batch and then immediately `MERGE`
relationships from existing parent nodes to that target.

PCG's canonical file write shape does exactly that:

```cypher
UNWIND $rows AS row
MERGE (f:File {path: row.path})
SET f += row.props
WITH f, row
MATCH (r:Repository {id: row.repo_id})
MERGE (r)-[:REPO_CONTAINS]->(f)
WITH f, row
MATCH (d:Directory {path: row.dir_path})
MERGE (d)-[:CONTAINS]->(f)
RETURN count(f)
```

Before this change, every relationship `MERGE` still called
`GetEdgeBetween(start, end, type)`. For a batch of file rows this meant two
relationship-existence probes per row, even when the `File` endpoint was just
created inside the same batch. On large repositories, those probes are costly
because they walk source-node relationship fanout that cannot contain an edge to
a node that did not exist before the current batch.

## Why this is safe

The optimization is scoped to one executor invocation and one batch-local fact:
if either endpoint was created earlier in this batch, no committed relationship
can already point to that endpoint. In that case the hot path can skip the
committed-store existence probe and create the relationship directly.

Duplicate input rows remain idempotent because the executor now keeps a
batch-local relationship cache keyed by `(startID, type, endID)`. Once a
relationship is created or found, later rows in the same batch reuse that cached
edge instead of creating duplicates.

## Changes

- Track node IDs created by `UnwindMergeChainBatch`.
- Cache relationship lookups and created relationships within the batch.
- Skip committed `GetEdgeBetween` probes when either relationship endpoint was
  created in the same batch.
- Document the batch-created endpoint behavior in the hot-path query cookbook.
- Add a regression test for the PCG file write shape, including duplicate input
  rows.
- Read the regression test's probe counter with `atomic.LoadInt64` so the test
  remains race-safe if the helper is reused under concurrent execution.

## Verification

- `go test -tags 'noui nolocalllm' ./pkg/cypher -run TestUnwindMergeBatch_SkipsRelationshipExistenceLookupForBatchCreatedEndpoint -count=1`
- `go test -tags 'noui nolocalllm' ./pkg/cypher ./pkg/storage -count=1`
- `git diff --check`

### DIFF
--- a/docs/performance/hot-path-query-cookbook.md
+++ b/docs/performance/hot-path-query-cookbook.md
@@ -576,6 +576,13 @@ idempotently. This shape is optimized by `UnwindMergeChainBatch`; it avoids the
 generic per-row fallback for `UNWIND ... MATCH ... MERGE ... SET += row.props
 ... MERGE relationship` ingestion pipelines.
 
+When the batch creates the relationship target and immediately links an existing
+parent to it, `UnwindMergeChainBatch` skips committed relationship existence
+probes for that batch-created endpoint and uses a batch-local relationship cache
+for duplicate rows. That keeps duplicate input idempotent while avoiding
+repeated scans across an already-large source-node fanout for relationships that
+could not have existed before the target node was created.
+
 ### 7.3e.2 Staged Compound `UNWIND` Version Pipeline
 
 ```cypher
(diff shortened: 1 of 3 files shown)


orneryd:
I appreciate the documentation update 🫶

linuxdynasty:
Addressed Copilot review in 5cc59e9. The regression test now reads edgeBetweenCalls with atomic.LoadInt64, matching the atomic increment and keeping the helper race-safe.

PR #125, opened 2026-04-28
Index exact relationship lookups

linuxdynasty:
## Why

PCG dogfood runs exposed a graph-size slope in idempotent relationship writes. The Cypher shape is ordinary `MERGE (start)-[:TYPE]->(end)`, but Badger-backed relationship lookup had to scan the start node outgoing edge index and then filter by end node and type. That means high-fanout nodes get slower as the graph fills, even when the write is logically an exact relationship-existence check.

This PR adds a direct edge-between lookup index so exact relationship checks can stay bounded by `(startID, endID, type)` instead of start-node out-degree.

## What changed

- Adds an edge-between Badger index keyed by start node, end node, relationship type, and edge ID.
- Maintains the index on edge create, update, delete, bulk, and transaction paths.
- Adds one-time startup backfill for existing stores using a marker key so older databases remain readable and automatically become indexed.
- Routes `GetEdgesBetween` / `GetEdgeBetween` through the direct index.
- Cleans the new prefix during database-prefix deletion.
- Documents the direct lookup path in the hot-path query cookbook.

## Verification

- `go test -tags 'noui nolocalllm' ./pkg/storage -run 'TestBadgerEngine_GetEdgeBetweenIndexMaintenance|TestBadgerEngine_EdgeBetweenIndexBackfill|TestBadgerEngine_GetEdgeBetween|TestKeyEncoding' -count=1`
- `go test -tags 'noui nolocalllm' ./pkg/storage -count=1`
- `go test -tags 'noui nolocalllm' ./pkg/storage ./pkg/cypher -count=1`
- `go build -tags 'noui nolocalllm' -o ./bin/nornicdb-headless-edge-index ./cmd/nornicdb`

I also tested the patched binary against a large stress repo from PCG dogfooding. The run drained healthy with `pending=0 in_flight=0 retrying=0 dead_letter=0 failed=0`. Canonical files completed 52 statements in 1.496s, Function completed 19,496 rows in 6.053s, Variable completed 118,768 rows in 62.340s, and max grouped execution stayed at 0.437s.

Follow-up medium-corpus proof also passed against a representative 23-repo corpus: projector finished 23/23, reducer finished 184 work items, and the queue ended at `pending=0 in_flight=0 retrying=0 dead_letter=0 failed=0`. In that run, the same large stress repo kept file relationship writes bounded: 52 file statements completed in 1.683s. Variable entity volume remained the tail at 69.750s, which is expected and separate from this relationship-existence fix.

### DIFF
--- a/docs/performance/hot-path-query-cookbook.md
+++ b/docs/performance/hot-path-query-cookbook.md
@@ -412,6 +412,11 @@ MERGE (a)-[r:LINKS_TO]->(b)
 SET r.updatedAt = $now;
 ```
 
+NornicDB maintains a direct relationship-existence index for this shape, keyed
+by start node, end node, relationship type, and edge ID. That keeps idempotent
+relationship `MERGE` checks from degrading with the start node's total outgoing
+degree.
+
 ### 6.4 Relationship Attach By ID
 
 ```cypher
--- a/docs/performance/hot-path-query-cookbook.md
+++ b/docs/performance/hot-path-query-cookbook.md
@@ -417,7 +417,7 @@ by start node, end node, relationship type, and edge ID. That keeps idempotent
 relationship `MERGE` checks from degrading with the start node's total outgoing
 degree.
 
-### 6.4 Relationship Attach By ID
+### 6.2 Relationship Attach By ID
 
 ```cypher
 MATCH (a:EntityA) WHERE elementId(a) = $fromId
@@ -427,7 +427,7 @@ CREATE (a)-[:LINKS_TO {createdAt: $now}]->(b);
 
 Use stable ID lookups before relationship creation for predictable attach latency.
 
-### 6.2 Bounded Traversal
+### 6.3 Bounded Traversal
 
 ```cypher
 MATCH p = (start:EntityA {primaryKey: $startKey})-[:LINKS_TO*1..3]->(n)
@@ -437,7 +437,7 @@ LIMIT 100;
 
 Always bound traversal depth and result count.
 
-### 6.3 Traversal With Early Filters
+### 6.4 Traversal With Early Filters
 
 ```cypher
 MATCH p = (start:EntityA {primaryKey: $startKey})-[:LINKS_TO*1..3]->(n:EntityB)
--- a/docs/performance/hot-path-query-cookbook.md
+++ b/docs/performance/hot-path-query-cookbook.md
@@ -412,10 +412,13 @@ MERGE (a)-[r:LINKS_TO]->(b)
 SET r.updatedAt = $now;
 ```
 
-NornicDB maintains a direct relationship-existence index for this shape, keyed
-by start node, end node, relationship type, and edge ID. That keeps idempotent
-relationship `MERGE` checks from degrading with the start node's total outgoing
-degree.
+NornicDB maintains direct relationship-existence indexes for this shape. The
+typed head key maps `(start node, end node, relationship type)` to one edge ID
+for the common single-edge lookup, while the set key keeps
+`(start node, end node, relationship type, edge ID)` entries for
+`GetEdgesBetween` and valid multiple same-type edges. Reads fall back to the
+legacy outgoing-edge scan and self-heal the new keys when opening mixed-version
+stores, so correctness does not depend solely on the one-time backfill marker.
 
 ### 6.2 Relationship Attach By ID
 
--- a/pkg/storage/badger.go
+++ b/pkg/storage/badger.go
@@ -18,22 +18,23 @@ import (
 // Key prefixes for BadgerDB storage organization
 // Using single-byte prefixes for efficiency
 const (
-	prefixNode          = byte(0x01) // nodes:nodeID -> Node
-	prefixEdge          = byte(0x02) // edges:edgeID -> Edge
-	prefixLabelIndex    = byte(0x03) // label:labelName:nodeID -> []byte{}
-	prefixOutgoingIndex = byte(0x04) // outgoing:nodeID:edgeID -> []byte{}
-	prefixIncomingIndex = byte(0x05) // incoming:nodeID:edgeID -> []byte{}
-	prefixEdgeTypeIndex = byte(0x06) // edgetype:type:edgeID -> []byte{} (for fast type lookups)
-	prefixPendingEmbed  = byte(0x07) // pending_embed:nodeID -> []byte{} (nodes needing embedding)
-	prefixEmbedding     = byte(0x08) // embedding:nodeID:chunkIndex -> []float32 (separate storage for large embeddings)
-	prefixSchema        = byte(0x09) // schema:global -> JSON(SchemaDefinition)
-	prefixTemporalIndex = byte(0x0A) // temporal:namespace:label:keyprops:keyhash:valid_from:nodeID -> []byte{}
-	prefixTemporalHead  = byte(0x0B) // temporal_current:namespace:label:keyprops:keyhash -> nodeID
-	prefixMVCCNode      = byte(0x0C) // mvcc_node:nodeID:version -> Node version payload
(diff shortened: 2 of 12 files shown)


orneryd:
One thing I am concerned about is the storage change

If I were shaping this for both performance and compatibility, I’d do this:

Add typed edge head key:
- edgeBetweenHead(start,end,type) -> edgeID
- Keep edgeBetweenSet(start,end,type,edgeID) -> empty only if multiple same-type edges are valid

On read:
- use head key for GetEdgeBetween
- use set prefix for GetEdgesBetween

Compatibility behavior:
- fallback to legacy scan on miss
- self-heal the new index from the fallback result
- Keep one-time backfill, but do not rely on the ready marker as the only correctness gate

That gives:
- fastest common-case lookup
- bounded behavior on high-fanout nodes
- no downgrade / mixed-version blind spot
- safer semver story

What do you think?

linuxdynasty:
I reworked this around the shape you suggested.

There is now a typed head key for the common `GetEdgeBetween(start,end,type)` path, while the set key stays in place for `GetEdgesBetween` and the valid “multiple same-type edges” case. If either new index is missing, reads fall back to the old outgoing-edge scan and repair the new keys from what they find, so mixed-version stores are still safe even if the ready marker says the backfill already ran.

I also updated create/update/delete/bulk/transaction maintenance, rebuilt both keyspaces during backfill, and added regression tests for self-heal plus same-type multi-edge behavior.

Verified with:
`go test -tags 'noui nolocalllm' ./pkg/storage ./pkg/cypher -count=1`

orneryd:
I have a feew more notes now that I am able to really look at it:

The implementation is sound, correctly maintains the index across all mutation paths (create, update, delete, bulk, cascade-on-node-delete, transactional buffered writes), and its compatibility story (fallback + self-heal + backfill marker) is well-designed. However, there's a few edge cases i think we need to iron out here before we pull it in. The two highest-impact improvements are bounding/deferring self-heal and making backfill non-blocking. 

| Priority | Item | File |
|---|---|---|
| High | Bound `selfHealEdgeBetweenIndexes` input slice; swallow `ErrConflict` | [pkg/storage/badger_queries.go](pkg/storage/badger_queries.go#L656) |
| High | Make `ensureEdgeBetweenIndex` async / context-aware / progress-logged | [pkg/storage/badger_edge_between_index.go](pkg/storage/badger_edge_between_index.go#L17) |
| Medium | Distinguish head-miss from edge-load-error in `edgeBetweenFromHeadIndex` | [pkg/storage/badger_queries.go](pkg/storage/badger_queries.go#L575) |
| Medium | Document/centralize `prefixMVCCMeta` sub-byte reservations | [pkg/storage/badger.go](pkg/storage/badger.go#L19) |
| Low | Use `errors.Is(err, badger.ErrKeyNotFound)` consistently | (multiple) |
| Low | Add audit-log line on backfill start/complete | [pkg/storage/badger_edge_between_index.go](pkg/storage/badger_edge_between_index.go) |
| Low | Snapshot/MVCC-read regression test for between-queries | [pkg/storage/badger_test.go](pkg/storage/badger_test.go) |



### High: Bound selfHealEdgeBetweenIndexes input slice; swallow ErrConflict in badger_queries.go:656.
Rationale: the current read path can turn a single legacy fallback into an unbounded write transaction that rewrites every discovered edge-between index entry. That is good for self-repair, but bad under load: a large result set can create write amplification, longer transaction windows, and extra contention in Badger’s conflict detection. Capping the number of healed edges per read keeps latency predictable and reduces the risk that a cold or partially upgraded store turns normal reads into heavy writes. Handling badger.ErrConflict as a non-fatal repair miss is also important because index healing is opportunistic, not correctness-critical once the fallback path already returned the right edge.

### High: Make ensureEdgeBetweenIndex async, context-aware, and progress-logged in badger_edge_between_index.go:17.
Rationale: backfilling by scanning every edge at engine startup is compatible, but it is operationally expensive on large stores. Right now it blocks open synchronously, provides no visibility into progress, and cannot be interrupted cleanly. That creates upgrade risk: the process may look hung, health checks may fail, and operators have no way to tell whether it is progressing or stalled. Making the rebuild cancellable and visible improves availability without changing the storage format. If done asynchronously, the existing legacy fallback path already provides a safe compatibility bridge while the index warms up.

### Medium: Distinguish head-miss from edge-load-error in edgeBetweenFromHeadIndex in badger_queries.go:575.
Rationale: a genuine “no head entry exists” condition is normal and should fall back quietly. A decode error or failed edge lookup is different: it may indicate stale secondary index state, data corruption, or a storage-layer issue. Returning nil for both cases hides the difference and makes it harder to detect integrity problems in production. Propagating the error, or at least logging it distinctly before falling back, would preserve debuggability and tighten operational safety without changing caller semantics much.

### Medium: Document or centralize prefixMVCCMeta sub-byte reservations in badger.go:19.
Rationale: the ready marker currently uses a raw {prefixMVCCMeta, 0x02} key next to other MVCC metadata keys. That works, but it creates a maintenance trap: future storage changes could accidentally reuse the same sub-byte because there is no obvious registry of reserved values. Centralizing those subkeys or moving index-specific markers under a dedicated prefix would reduce collision risk and make the storage layout easier to extend safely over time.

### Low: Use errors.Is(err, badger.ErrKeyNotFound) consistently in badger_queries.go:580.
Rationale: direct equality checks are fine if the dependency always returns the bare sentinel, but errors.Is is more robust and idiomatic. It makes the code resilient if errors are wrapped in the future and improves consistency across the storage layer. This is a low-priority cleanup because it is unlikely to change behavior today, but it hardens the code against subtle compatibility regressions.

### Low: Add audit-log entries for backfill start and completion in badger_edge_between_index.go:17.
Rationale: rebuilding a secondary index is an operationally meaningful event. It can affect startup time, disk I/O, and write pressure, especially after upgrades or backup restores. Without logging, operators have no durable record that a rebuild occurred or how long it took. Adding audit or structured operational logging would make support easier and reduce ambiguity when diagnosing slow boots or unexpected write spikes.

### Low: Add a snapshot/MVCC-read regression test for between-queries in badger_test.go.
Rationale: the new index paths optimize latest-state reads, but storage correctness also depends on not breaking snapshot semantics if historical or MVCC-aware read paths rely on edge lookup behavior. Even though the current package tests pass, a targeted regression test would confirm that the new direct index does not accidentally bypass versioned expectations or introduce differences between indexed reads and legacy fallback reads. This is lower priority because there is no immediate evidence of a bug, but it is the right place to lock in compatibility.

orneryd (changes requested):
see above comment

linuxdynasty:
I tightened this up around the upgrade-safety concerns you called out.

The read-triggered self-heal is now capped, and `badger.ErrConflict` during that repair path is treated as a skipped opportunistic repair instead of a read failure. Startup backfill also no longer blocks opening a non-empty upgraded store: empty stores mark the index ready immediately, while non-empty stores start a cancellable background rebuild with start, progress, completion, cancel, and failure logs. `Close` now cancels and waits for that rebuild before closing Badger, so we do not leave a goroutine writing against a closed DB.

I also split a normal head-index miss from an edge-load/decode problem, centralized the MVCC metadata subkey used by the ready marker, switched the touched key-not-found checks to `errors.Is`, and added MVCC snapshot coverage so latest-state edge-between indexes do not leak into historical between reads.

Verified with:
`go test -tags 'noui nolocalllm' ./pkg/storage ./pkg/cypher -count=1`

I also rebuilt the branch and ran a PCG one-repo `local_authoritative` smoke against it. That completed cleanly in 76s: 1 projector item succeeded, all 8 reducer domains succeeded, SQL relationships wrote 286 edges, code-call projection wrote 9,158 rows, and there were no retry/dead-letter/failed queue rows.

orneryd:
Looks good thanks!

PR #126, opened 2026-04-29
Make NornicDB observable: metrics, traces, logs, and K8s integration

linuxdynasty:
**This is the M1 PR.** Per @orneryd, the entire 13-phase observability milestone ships as a single PR on the `otel` branch — no per-phase PRs. This PR stays open and grows with each phase; final merge is at M1 completion (post-Phase 12). Phase audit trails live in ADR-0001 §4.1 as commit-range references rather than PR links.

## What we're doing

NornicDB doesn't yet give operators the visibility they expect from a production database. There's no way to set up SLOs, trace a slow query through Bolt → Cypher → storage, see why replication is lagging, or detect that a tenant is generating a cardinality bomb in metric labels. Customers running NornicDB on Kubernetes hit this immediately — the standard `kube-prometheus-stack` flow doesn't work, there's no `ServiceMonitor` template, and there's no tracing pipeline at all.

This PR is the work to fix that. We're adding a `pkg/observability` package that becomes the single seam for metrics, traces, and structured logs across the codebase, plus a dedicated unauthenticated telemetry port (`:9090`) that Prometheus scrapes directly with no bearer-token plumbing. The same numbers also stream out via OTLP push for customers who already run a collector. A bundled Helm chart ships `ServiceMonitor`, `PodMonitor`, `NetworkPolicy`, alert rules, and a Grafana dashboard so the K8s flow works on first install with no config.

The current contents of this PR are Phase 0 — the design contract for everything above (ADR-0001). Subsequent commits on `otel` will deliver Phases 1 through 12 (the actual implementation, ~13 logical chunks of work). The PR will keep growing.

## What we'll collect, and why

**Metrics** — about 60 metric families across nine subsystems (HTTP, Bolt, Cypher, Storage/Badger, MVCC, Embeddings, Search, Replication, Auth, plus Cache and Go runtime). Modeled on `pg_stat_*` from Postgres and `db.*` JMX MBeans from Neo4j, but renamed to NornicDB semantics. Hand-instrumented via `prometheus/client_golang` with hot-path label binding pre-bound at construction so the observation cost stays under 2 heap allocs per call. Exposed at `:9090/metrics` for pull and OTLP-bridged in parallel for push customers.

The notable design constraints baked in:
- **Naming discipline.** Every counter ends in `_total`, every histogram ends in the unit it measures, three named bucket constants are the only allowed bucket types. A registration helper rejects metrics that violate the rules at startup so it's impossible to ship a histogram with arbitrary buckets.
- **Forbidden labels.** Full HTTP path, raw Cypher query, user/IP/UUID, embedding text, and span/trace IDs are all on a runtime allow-list — registering a metric with one of these labels fails at init. Cardinality bombs are the most common observability failure mode and we want them to be impossible by construction, not just discouraged in code review.
- **Tenant kill switch.** A `metrics.tenant_labels_enabled` flag (default-on for K8s, default-off otherwise) controls whether the `database` label appears on tenant-scoped subsystems (`nornicdb_http_*`, `nornicdb_storage_*`, `nornicdb_cypher_*`, `nornicdb_search_*`, `nornicdb_mvcc_*`). Operators who run on shared infrastructure but don't want tenant names leaving the pod can flip it without losing other metrics.
- **Two egress paths.** Prometheus pull is canonical at `:9090/metrics`. OTel meter readings stream the same numbers in parallel through the `otelprom` bridge under a `nornicdb_otel_*` namespace so they don't collide with hand-instrumented `nornicdb_*` series.
- **Backwards compat.** The legacy `:7474/metrics` endpoint stays enabled in M1 with `Deprecation: true` and `Sunset: <date>` headers. Removal is a separate v1.1.0 PR alongside #117 — explicitly out of scope for this M1 work.

**Traces** — OpenTelemetry SDK with OTLP/gRPC primary, OTLP/HTTP fallback. The default sampler is `TraceIDRatioBased(0.01)` standalone — it does *not* honor parent context by default, so an upstream service running 100% sampling can't blow up our trace volume. Two opt-in modes: `parent_capped` honors upstream sampling up to a configurable QPS ceiling (default 100/s), and `parent_strict` does full upstream honor with a startup `WARN` about volume risk. Spans cover the HTTP edge, Bolt session and per-message dispatch, Cypher executor (with per-logical-operator child spans), storage adapter, embedding worker, search service, replication append/apply paths, and gRPC inter-service traffic via `otelgrpc` interceptors. The replication codec gets a `codec_version` field added before the optional `traceparent` field, in two separate commits on `otel`, so rolling upgrades don't break.

The hot-path overhead budget is ≤2% on `make bench-cypher` (tracing on or off) and ≤5% on `make bench-bolt`. BSP tuning is locked at `MaxQueueSize=8192`, `MaxExportBatchSize=1024`, `BatchTimeout=2s` — production-correct for bursty DB workloads, with self-instrumentation (`bsp_queue_depth` gauge, `bsp_dropped_spans_total` counter) so operators can see when the pipeline is saturated. OTLP TLS is required in production; plaintext is rejected unless the standard `OTEL_EXPORTER_OTLP_INSECURE=true` is set explicitly.

**Structured logs** — the ~165 `log.Printf` and `fmt.Println` sites across `pkg/server`, `pkg/cypher`, `pkg/storage`, and `pkg/bolt` migrate to the stdlib `log/slog` interface. Every record carries `time` (RFC3339Nano), `level`, `msg`, `service=nornicdb`, `version`, `node_id`, plus `trace_id`/`span_id` when there's an active span. A PII redactor sharing the audit allow-list (`password`, `token`, `authorization`, `secret`, `api_key`) strips matching keys at any nesting depth. The slow-query log emits truncated 500-char queries with literal redaction (`<REDACTED>` for string/number literals before truncation) so the full query text never appears as a metric label or span attribute. Call sites code against the slog interface; the underlying handler implementation is a Phase 2 perf-benchmark decision (stdlib `slog.NewJSONHandler` first, custom slog-compatible handler if it doesn't meet the per-record allocation budget). `pkg/audit/audit.go` is *not* migrated — audit stays a distinct, signed, append-only stream with its own retention policy.

**K8s integration** — the bundled Helm chart ships `ServiceMonitor` (default-on) and `PodMonitor` (opt-in) targeting the new `:9090` port, a `NetworkPolicy` that defaults to monitoring-namespace-only ingress, a `startupProbe` with high `failureThreshold` (≈10 min) for storage warm-up, and a `preStop` lifecycle hook that sleeps 10s before SIGTERM to drain `kube-proxy` endpoints. `/readyz` returns 200 throughout the lifecycle with a progress JSON body so kubelet treats the pod as routable as soon as core surfaces are up while still surfacing warm-up state. Reference Grafana dashboards and versioned alert rules ship at `docs/operations/`. A CI integration test against `kube-prometheus-stack` plus Tempo confirms exemplar correlation, NetworkPolicy enforcement, scrape behavior, and OTel→Prom bridge parity end-to-end.

## Why this matters

The honest version: today, when a customer reports that NornicDB is slow under load on their cluster, we have no way to tell them *which* layer is slow. Is it Bolt protocol decode? Cypher planning? A storage compaction stall? Replication lag? Embedding queue backpressure? We can guess, but we can't show them. After this work, they can.

The same is true for security posture. Right now there's no way to detect a brute-force auth attempt as a metric, no way to alert on tenant-name enumeration via the public metrics endpoint, and no way to prove to a compliance auditor that PII isn't leaking through observability. After this, those things have specific metric names, specific recording rules, and specific test coverage.

And it has to be installable on K8s without a `values.yaml` workshop — the goal is `helm install` and have Prometheus scrape, alerts fire, dashboards work, and tenant labels behave correctly all on first try. Anything less is operational debt we don't want to ship with.

## Reviewer setup

Two humans, four roles. The roster is in project memory:

- **orneryd** — Architecture + Public-API owner. Approved the ADR design (Phase 0) on 2026-04-30. Subsequent commits will land on this PR for Phases 1–12; orneryd's review may be auto-dismissed by GitHub when those commits land — that's expected. The §5 sign-off table inside the ADR is the durable audit trail.
- **linuxdynasty** (PR author) — SRE/Ops + Security. GitHub doesn't let PR authors approve their own PRs, so these two role rows in §5 are filled in as self-attestation with the same PR URL + date.

## Audit trail

The §4.1 "Phase exit references" table inside the ADR is the GOV-03 audit trail. It has a row per phase (0 through 12) with three columns:

- `Commit range` — the inclusive SHA range on `otel` that delivered that phase
- `Verified` — the date the verifier signed off
- `Verifier` — the GitHub handle of whoever ran the verifier

Phase 0's row gets filled in once the finalize commit lands (Status flip + §5 fill + §4.1 row 0). Subsequent phases append their own rows as they exit.

## What's next on this PR

- A finalize commit is queued: flip ADR `Status: **Proposed**` → `**Accepted**`, fill §5 reviewer rows with orneryd + linuxdynasty + 2026-04-30 + this PR's URL, fill §4.1 row 0 with the Phase 0 commit range. **Does not merge** — PR stays open.
- Phase 1 (Observability Foundation Skeleton) starts after that. Commits land on `otel`. New `pkg/observability` package skeleton.
- Phases 2–12 follow, one at a time, on the same branch.
- Final review + merge to `main` happens after Phase 12 exits.


docs/architecture/adr/0001-observability.md
```diff
@@ -0,0 +1,666 @@
...
+    # OTel-standard env vars are honored: OTEL_EXPORTER_OTLP_ENDPOINT, etc.
+  logging:
+    level: "INFO"                       # NORNICDB_LOG_LEVEL
+    format: "json"                      # NORNICDB_LOG_FORMAT (default flips to json)
+    output: "stdout"                    # NORNICDB_LOG_OUTPUT
+    redact_extra: []                    # NORNICDB_LOG_REDACT_EXTRA (csv)
+  pprof:
+    enabled: false                      # NORNICDB_PPROF_ENABLED
+    listen: ":9091"
+```
```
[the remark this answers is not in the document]
linuxdynasty: Fixed in 7c9fe78 — the config example now reads `listen: "127.0.0.1:9091"` so the YAML matches the loopback-by-default rule stated two paragraphs up. Marking resolved.

docs/architecture/adr/0001-observability.md
```diff
@@ -0,0 +1,668 @@
...
+  `pkg/audit` to keep the rules in one place.
+* The slow-query log emits `WARN` with the truncated 500-char query, the
+  `plan_hash`, and `cypher.duration_ms`. The full query text is **never**
+  attached as a metric label or span attribute — it goes to the log only.
+* `pkg/audit/audit.go` is **not** migrated to `slog`. Audit must remain a
+  distinct, signed, append-only stream with its own retention policy.
+
+### 2.6 Kubernetes service monitor — dedicated unauthenticated port
+
+A new listener `:9090` (configurable via `NORNICDB_TELEMETRY_LISTEN`) serves:
```
orneryd: NORNICDB_TELEMETRY_PORT={number_without_colon}

docs/architecture/adr/0001-observability.md
```diff
@@ -0,0 +1,668 @@
...
+### 2.7 Configuration surface
+
+New `Observability` block in `pkg/config/config.go`, all overridable via
+environment so containers stay declarative:
+
+```yaml
+observability:
+  metrics:
+    enabled: true                       # NORNICDB_METRICS_ENABLED
+    listen: ":9090"                     # NORNICDB_TELEMETRY_LISTEN
```
orneryd: same as above

docs/architecture/adr/0001-observability.md
```diff
@@ -0,0 +1,668 @@
...
+    legacy_data_plane_endpoint: true    # NORNICDB_LEGACY_METRICS  (deprecated)
```
orneryd: I don't think we need a flag for this. just leave it enabled until we remove it/change it.

docs/architecture/adr/0001-observability.md
```diff
@@ -0,0 +1,668 @@
...
+This keeps the Prometheus and OTel SDK surface out of business logic, makes
+testing straightforward (interface seam), and lets us swap exporters
+without a tree-wide diff.
+
+```
+pkg/observability/
+  observability.go       // Init(ctx, cfg) -> Provider; Shutdown(ctx)
+  metrics.go             // package-level registry + helpers (must call Init first)
+  tracing.go             // tracer, propagators, span helpers
+  logging.go             // slog handler, redactor, ctx-aware logger
```
orneryd, line 561: im inclined to do a little research, if we take a dependency, I want to make sure it's high-performance. if we need to write our own high-performance logger I am not above doing that.

docs/architecture/adr/0001-observability.md
```diff
@@ -0,0 +1,668 @@
...
+  middleware_http.go     // wraps mux: otelhttp + per-route histogram
+  middleware_bolt.go     // hooks into pkg/bolt server lifecycle
+  exemplar.go            // histogram exemplar wiring (trace_id from ctx)
```
orneryd, line 568: grpc metrics?

docs/architecture/adr/0001-observability.md
```diff
@@ -0,0 +1,668 @@
...
+  between us and a Prometheus blow-up. Code review must enforce it; a
+  golangci-lint custom analyzer is a follow-up.
+
+### 3.3 Risks and mitigations
+
+| Risk                                                                               | Mitigation                                                                                                  |
+|------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------|
+| Hot-path overhead from span creation in Cypher executor.                           | Sample at the trace level, not the span level; pre-bind counters; benchmark gate ≤2%.                        |
+| OTLP endpoint misconfiguration crashes the server on startup.                      | Init returns a `noop` provider on exporter error and logs `WARN`; never fatal.                              |
+| Customer scrapers depending on the legacy `:7474/metrics` break on removal.        | Two-phase deprecation with `Sunset` header, release notes, and metric-name parity on `:9090`.                |
```
orneryd: definitetly a removal i would do with 1.1.0 to go along with #117

docs/architecture/adr/0001-observability.md
```diff
@@ -0,0 +1,668 @@
...
+| KD-05 | Default sampler should be `ParentBased` to honor upstream sampling decisions               | Reject — storage layer must control its own trace volume; default is `TraceIDRatioBased(0.01)` standalone. `ParentBased`-style behavior is opt-in only.   | §2.4 (Sampler — default + opt-in `parent_capped` / `parent_strict` modes)                         |
+| KD-06 | NetworkPolicy alone is sufficient defense for tenant-name leak through `:9090` labels      | Reject — non-K8s deployments lack NetworkPolicy. Add `metrics.tenant_labels_enabled` flag with K8s auto-detect (default `true` on K8s, `false` otherwise). | §2.2 (label kill-switch) and §2.6 (K8s detect signals + override)                                 |
+| KD-07 | Cut over `:7474/metrics` to the new schema in one release                                  | Reject — published metric names are a customer contract. Phase 1 ships a translation layer that emits legacy names from the new registry; one-cycle deprecation overlap with `Sunset` header. | §2.6 (legacy data-plane endpoint + Deprecation/Sunset headers)                                    |
```
orneryd, line 678: I just want to make sure that for this feature we leave the existing endpoints alone with just deprecation markers, removal i'll open a separate issue/PR for

docs/architecture/adr/0001-observability.md
linuxdynasty: Done in 2f451f4 — `NORNICDB_TELEMETRY_PORT=9090` (just a port number, no colon). Bind address stays `:` (all interfaces) and the bundled `NetworkPolicy` restricts ingress to the monitoring namespace; if an operator needs to lock the bind address itself they layer a custom NetworkPolicy on top.

docs/architecture/adr/0001-observability.md
linuxdynasty: Same fix in 2f451f4 — YAML now reads `port: 9090  # NORNICDB_TELEMETRY_PORT`.

docs/architecture/adr/0001-observability.md
linuxdynasty: Dropped in 2f451f4. The `legacy_data_plane_endpoint` flag is gone; `:7474/metrics` stays enabled in M1 unconditionally. Removal is post-M1 (see the §3.3 risk row + #117).

docs/architecture/adr/0001-observability.md
linuxdynasty: Reframed in 2f451f4. §2.5 now codes against the `slog` *interface* — handler implementation is a Phase 2 perf-benchmark decision. If stdlib `slog.NewJSONHandler` doesn't meet the per-record allocation budget on hot paths, a custom slog-compatible `Handler` is fair game. The decision (and benchmark numbers) get recorded in the Phase 2 SUMMARY.md so we have an audit trail.

docs/architecture/adr/0001-observability.md
linuxdynasty: Added in 2f451f4. §2.8 implementation map now includes `middleware_grpc.go` — `otelgrpc` Unary + Stream interceptors covering replication peer-to-peer traffic and any future gRPC surfaces. Standard gRPC server/client metrics (`rpc.server.duration`, `rpc.server.requests`, etc.) come for free from the contrib instrumentation. The replication subsystem metrics in §2.3 (`lag_bytes`, `rtt_seconds`, etc.) sit on top of those raw gRPC numbers.

docs/architecture/adr/0001-observability.md
linuxdynasty: Updated in 2f451f4 — the §3.3 risk row now reads: removal scheduled for v1.1.0 in a separate follow-up PR alongside #117, customers get at least one minor-release overlap window.

docs/architecture/adr/0001-observability.md
linuxdynasty: Confirmed and made unambiguous in 2f451f4. §2.6 prose now explicitly says **M1 ships only deprecation markers — no removal logic**, with the actual removal scheduled for v1.1.0 in a separate PR alongside #117 (explicitly out of scope for ADR-0001). I left them unresolved so you can be the one to resolve once it reads right to you.

docs/architecture/adr/0001-observability.md
```diff
@@ -0,0 +1,699 @@
...
+* `pkg/audit/audit.go` is **not** migrated to `slog`. Audit must remain a
+  distinct, signed, append-only stream with its own retention policy.
+
+### 2.6 Kubernetes service monitor — dedicated unauthenticated port
+
+A new listener on port `9090` (configurable via `NORNICDB_TELEMETRY_PORT`, value is the port number with no colon prefix; bind address is `:` so the listener accepts on all interfaces and the NetworkPolicy below restricts ingress) serves:
+
+| Path           | Auth      | Purpose                                                              |
+|----------------|-----------|----------------------------------------------------------------------|
+| `/metrics`     | **None**  | Prometheus pull endpoint. Returns `client_golang` registry.          |
```
orneryd, line 459: need to define RBAC entitlements and assign default roles for access, with a flag like NORNICDB_TELEMETRY_AUTH=true (false by default) and assign all the entitlements to the builtin admin role. even if for now a single role "telemetry" is defined that would at least allow for high-security environments that need to access these metrics.

we hook into the existing authentication mechanism for these endpoints.

docs/architecture/adr/0001-observability.md
```diff
@@ -0,0 +1,699 @@
...
+The opt-in `:9091` pprof listener binds **`127.0.0.1` by default** (not `:9091` interpreted as `0.0.0.0:9091`); binding all interfaces would make pprof a credential-exfiltration surface. Operators who need remote pprof access must explicitly configure a non-loopback listen address and a corresponding NetworkPolicy.
+
+**Why a separate port and not just "remove auth from `/metrics`":**
+
+* Metric labels include `database` (tenant name). Exposing
+  `nornicdb_storage_nodes_total{database="acme-prod"}` to anyone who can
+  reach `:7474` from the public internet is a tenant-naming oracle. The
+  defense is *network-scoped*, not bearer-scoped — exactly how Postgres'
+  `postgres_exporter` and CockroachDB's `_status/vars` do it.
+* Prometheus Operator's `ServiceMonitor` does not natively send bearer
```
orneryd, line 482: the assertion here should be updated that even if Prometheus doesn't send the `Authentication` header, we should still accept it and handle that flow correctly, with the runtime flag set wether or not this is wide open or authenticated access.

orneryd: Note: if we hook in through the existing authentication system we would technically support oauth for free here too.

orneryd:
merged with #153

PR #169, opened 2026-05-16
fix(mcp,heimdall): return cosine similarity from discover, not RRF rank score

41nathand:
## Summary

- `discover`'s `similarity` field was the outer RRF rank-decay score (`1/(60+rank)`, capped at ~0.0164 at rank 1, decaying to ~0.0091 at rank 50). Because that value depends only on rank position, every query returned an identical top-to-bottom score sequence regardless of content, making the `min_similarity` parameter useless.
- This PR makes both fusion paths (`pkg/mcp/server.go:handleDiscover` and `pkg/heimdall/metrics.go:Discover`) surface the strongest underlying similarity across chunks (cosine for vector hits, BM25 raw for lexical-only hits), while keeping outer RRF for sort order so multi-chunk fusion stability is preserved.
- The BM25-only fallback in `pkg/mcp` also previously surfaced `r.Score` (raw BM25) as `similarity`; it now uses `r.Similarity`, keeping the field's semantics consistent across code paths.
- Adds two targeted regression tests that fail under the old behaviour and pass under the new.

## Reproducer (before fix)

```js
// MCP discover via the tool surface
const r1 = await discover({ query: 'completely random nonsense', limit: 5, min_similarity: 0 })
const r2 = await discover({ query: 'Diego is 3 years old',       limit: 5, min_similarity: 0 })
console.log(r1.results.map(r => r.similarity))
// [0.01639, 0.01613, 0.01587, 0.01562, 0.01538]
console.log(r2.results.map(r => r.similarity))
// [0.01639, 0.01613, 0.01587, 0.01562, 0.01538]   ← identical sequence
```

`r2`'s _IDs_ are query-correct (vector search works) — but the `similarity` numbers are the rank-decay function `1/(60+k)`, not cosine. Any caller using `min_similarity > 0.02` silently filters everything, including the topical match.

## Root cause

`fuseRRF` already preserves the underlying cosine score in `search.SearchResult.Similarity`. Both fusion paths (per-chunk RRF) then overwrote it with the outer rank score:

- `pkg/mcp/server.go` — `Similarity: f.scoreRRF` on the result
- `pkg/heimdall/metrics.go` — `Score: f.scoreRRF` on the intermediate `SemanticSearchResult`, then propagated as `Similarity: r.Score`

## Fix

Each fused entry now tracks `bestSim` (max underlying `r.Similarity`/`r.Score` across chunks). That value is surfaced as the result's similarity and is what `min_similarity` filters against. Outer RRF still drives sort order — sort stability across multi-chunk fusion is what RRF is for.

```go
// pkg/mcp/server.go (handleDiscover, vector path)
type fused struct {
    ...
    scoreRRF float64 // outer RRF used for ordering across chunks
    bestSim  float64 // max underlying similarity (cosine or BM25 raw)
}

f.scoreRRF += 1.0 / (outerRRFK + float64(rank+1))
if r.Similarity > f.bestSim {
    f.bestSim = r.Similarity
}
...
if minScore > 0 && f.bestSim < minScore { continue }
...
res := SearchResult{..., Similarity: f.bestSim}
```

Same structural change in `pkg/heimdall/metrics.go`.

The BM25-only fallback in `pkg/mcp` now reads `r.Similarity` rather than `r.Score`, so raw BM25 fused scores don't leak into the same field that elsewhere carries cosine.

## Behaviour after fix

```
discover('Diego son age')    →  top similarity 0.66–0.73   (cosine; topical match)
discover('Nathan')           →  top similarity 5–8         (BM25-dominant, exact-string hits)
discover('orthogonal query') →  results have low cosine values OR drop out at min_sim ≥ 0.3
```

`similarity` is now in two regimes depending on which retrieval pathway dominated for a given node: cosine `[0, 1]` for vector-pipeline hits, raw BM25 (typically 1–40) for BM25-only hits. Any threshold ≥ ~0.3 still filters cosine noise; lexical-only hits effectively bypass the threshold by design (a strong exact-string match is rarely something the caller wants to filter out).

## Tests

`pkg/mcp/server_test.go`:
- `TestHandleDiscover_SimilarityIsCosineNotRRF` — sets up a `match` node whose embedding equals the query embedding and a `miss` node with an orthogonal embedding. Asserts (1) the match surfaces with cosine ~1.0, not ~0.0164; (2) `min_similarity=0.5` drops the orthogonal miss but keeps the match.

`pkg/heimdall/metrics_discover_chunk_test.go`:
- `TestQueryExecutor_Discover_SimilarityIsBestScore` — stages two chunk batches scoring the same node at 0.55 and 0.92 across the multi-chunk fusion path. Asserts the surfaced `Similarity` is 0.92 (max across chunks), not the cumulative outer-RRF score (~0.0328) the old code produced.

Both tests have been confirmed to fail without the fix and pass with it.

A small enabling change is made to `mockEmbedder.EmbedBatch` so it honours the configurable `embedding` field (previously only `Embed` did). This keeps the existing default (zero vectors) unchanged for prior tests.

## Test plan

- [x] `go test ./pkg/mcp/` (passes including new regression test)
- [x] `go test ./pkg/heimdall/` (passes including new regression test)
- [x] Reverted the fix locally and confirmed both new tests fail at the documented values (0.01639 for mcp, 0.03279 for heimdall — exactly the RRF rank-decay values)
- [x] Probe against live local NornicDB after rebuild: unrelated queries no longer share a top-score sequence; topical queries surface cosine ~0.7 for relevant hits

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/heimdall/metrics.go
+++ b/pkg/heimdall/metrics.go
@@ -443,7 +443,8 @@ func (e *QueryExecutor) Discover(ctx context.Context, query string, nodeTypes []
 
 			type fused struct {
 				best     *SemanticSearchResult
-				scoreRRF float64
+				scoreRRF float64 // outer RRF used for ordering across chunks
+				bestSim  float64 // strongest underlying similarity across chunks
 			}
 			fusedByID := make(map[string]*fused)
 
@@ -466,12 +467,13 @@ func (e *QueryExecutor) Discover(ctx context.Context, query string, nodeTypes []
 					}
 					f := fusedByID[r.ID]
 					if f == nil {
-						f = &fused{best: r}
+						f = &fused{best: r, bestSim: r.Score}
 						fusedByID[r.ID] = f
 					}
-					// Outer RRF: 1/(k + rank), rank is 1-based.
+					// Outer RRF: 1/(k + rank), rank is 1-based. Used for ordering only.
 					f.scoreRRF += 1.0 / (outerRRFK + float64(rank+1))
-					if f.best == nil || r.Score > f.best.Score {
+					if r.Score > f.bestSim {
+						f.bestSim = r.Score
 						f.best = r
 					}
 				}
@@ -498,11 +500,15 @@ func (e *QueryExecutor) Discover(ctx context.Context, query string, nodeTypes []
 					if f.best == nil {
 						continue
 					}
+					// Surface the strongest underlying similarity (cosine when
+					// available), NOT the outer RRF rank-derived score. This
+					// restores meaningful min_similarity threshold semantics
+					// downstream.
 					searchResults = append(searchResults, &SemanticSearchResult{
 						ID:         f.best.ID,
 						Labels:     f.best.Labels,
 						Properties: f.best.Properties,
-						Score:      f.scoreRRF,
+						Score:      f.bestSim,
 					})
 				}
 			}
--- a/pkg/heimdall/metrics_discover_chunk_test.go
+++ b/pkg/heimdall/metrics_discover_chunk_test.go
@@ -123,6 +123,82 @@ func loadLargeDocQuery(t *testing.T) string {
 	return query
 }
 
+// TestQueryExecutor_Discover_SimilarityIsBestScore is a regression test for the
+// scoring bug where the outer RRF rank-decay score (`1/(60+rank)`, capped at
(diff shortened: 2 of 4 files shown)


orneryd:
thanks dude i'll take a look in the morning.m!

PR #230, opened 2026-07-01
Speed up bound relationship deletes

linuxdynasty:
## Summary

This PR improves one narrow relationship-delete shape: a query first binds a source node, then deletes one matching relationship from that source.

```cypher
MATCH (source:Label {id: ...})
MATCH (source)-[rel:TYPE|OTHER_TYPE]->(:TargetLabel)
WHERE rel.some_property = ...
DELETE rel
```

For that form, the executor can start from the already-bound source node instead of planning a broad traversal and discarding unrelated relationships later. The fast path still applies the relationship type, relationship properties, target-node pattern, and `WHERE` clause before deleting anything. Queries outside this strict shape continue through the existing delete paths.

The review pass tightened the safety boundaries around that optimization:

- The fast path now rejects non-simple row pipelines unless the only interposed pipeline is `WITH <rel> LIMIT <positive integer>`. It rejects `ORDER BY`, `SKIP`, `CALL`, `UNWIND`, and `OPTIONAL MATCH` in this path.
- `WITH rel LIMIT n DELETE rel` is now handled explicitly and deletes at most the limited row count.
- Transaction reads for `AllNodes`, `GetOutgoingEdges`, and `GetIncomingEdges` now use MVCC-visible committed state instead of current engine heads/indexes.
- Inside transaction wrappers, source-node collection avoids current property indexes and filters a snapshot-visible label/all-node population instead.
- Storage lookup errors from candidate relationship and target-node reads are returned to the caller instead of being treated as empty matches.

One deliberate trade-off: transaction-wrapper adjacency reads now prefer correctness over the current adjacency index until there is a snapshot-aware adjacency index. Non-transactional reads still use the existing adjacency index path.

## Verification

Passed locally on the rebased branch (`c4901451`):

- `go test ./pkg/cypher -run 'TestMatchPatternProperty_(BoundRelationshipDeleteSegmentEligibility|TryBoundRelationshipDeleteDeletesSimpleMatch|TryBoundRelationshipDeleteDeletesUnionAndLimitedMatches|ExecuteDeleteUsesBoundRelationshipDelete|BoundRelationshipDeleteRoutingPredicates|BoundRelationshipDeleteHonorsWithLimit|BoundRelationshipDeleteUsesTransactionSnapshot|BoundRelationshipDeleteUsesAdjacency|BoundRelationshipDeleteReturnsCandidateLookupErrors|TryBoundRelationshipDeletePropagatesCandidateLookupError|BoundRelationshipDeleteReturnsTargetLookupError)' -count=1 -v`\n- `go test ./pkg/storage -run 'TestTransaction_(AllNodesRepeatabilityWithinSnapshot|AdjacencyRepeatabilityWithinSnapshot|LabelScanRepeatabilityWithinSnapshot|EdgeTraversalRemainsSnapshotConsistentAcrossConcurrentDelete)' -count=1 -v`\n- `go test ./pkg/cypher -count=1`\n- `go test ./pkg/storage -count=1`\n- `go test ./pkg/cypher -run '^TestBenchmarkMatchCreateDelete$' -count=1 -v` — local run: `43341.64 ops/sec`, `0.023 ms/op`\n- `git diff --check`\n\nLocal setup note: `./scripts/build-llama.sh b9835` was run first so the normal untagged test binary links against `lib/llama/libllama_darwin_arm64.a`.\n

### DIFF
--- a/pkg/storage/badger_transaction.go
+++ b/pkg/storage/badger_transaction.go
@@ -1382,7 +1382,7 @@ func (tx *BadgerTransaction) GetOutgoingEdges(nodeID NodeID) ([]*Edge, error) {
 		return nil, err
 	}
 
-	committed, err := tx.getCommittedAdjacentEdgesLocked(nodeID, "outgoing")
+	committed, err := tx.getCommittedAdjacentEdgesLocked(nodeID, Outgoing)
 	if err != nil {
 		return nil, err
 	}
@@ -1402,7 +1402,7 @@ func (tx *BadgerTransaction) GetIncomingEdges(nodeID NodeID) ([]*Edge, error) {
 		return nil, err
 	}
 
-	committed, err := tx.getCommittedAdjacentEdgesLocked(nodeID, "incoming")
+	committed, err := tx.getCommittedAdjacentEdgesLocked(nodeID, Incoming)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/storage/badger_transaction.go
+++ b/pkg/storage/badger_transaction.go
@@ -1382,7 +1382,7 @@ func (tx *BadgerTransaction) GetOutgoingEdges(nodeID NodeID) ([]*Edge, error) {
 		return nil, err
 	}
 
-	committed, err := tx.engine.GetOutgoingEdges(nodeID)
+	committed, err := tx.getCommittedAdjacentEdgesLocked(nodeID, "outgoing")
 	if err != nil {
 		return nil, err
 	}
@@ -1402,7 +1402,7 @@ func (tx *BadgerTransaction) GetIncomingEdges(nodeID NodeID) ([]*Edge, error) {
 		return nil, err
 	}
 
-	committed, err := tx.engine.GetIncomingEdges(nodeID)
+	committed, err := tx.getCommittedAdjacentEdgesLocked(nodeID, "incoming")
 	if err != nil {
 		return nil, err
 	}
@@ -1510,7 +1510,7 @@ func (tx *BadgerTransaction) AllNodes() ([]*Node, error) {
 		return nil, err
 	}
 
-	committed, err := tx.engine.AllNodes()
+	committed, err := tx.getAllCommittedNodesLocked()
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/cypher/executor_mutations.go
+++ b/pkg/cypher/executor_mutations.go
@@ -155,6 +155,12 @@ func (e *StorageExecutor) executeDelete(ctx context.Context, cypher string) (*Ex
 	// LIMIT/SKIP/WITH/ORDER BY/CALL/UNWIND in the MATCH segment.
 	matchSegment := strings.TrimSpace(cypher[matchIdx:deleteIdx])
 	_, inTransactionWrapper := e.getStorage(ctx).(*transactionStorageWrapper)
+	if hot, ok, err := e.tryExecuteBoundRelationshipDelete(ctx, matchSegment, cypher, deleteVars, detach); ok || err != nil {
+		if err != nil {
+			return nil, err
+		}
+		return hot, nil
+	}
 	if !inTransactionWrapper && e.isDeleteStreamingEligible(matchSegment, deleteVars, detach) {
 		result, err := e.executeDeleteStreaming(ctx, matchSegment, deleteVars, needEdgeStats)
 		if err != nil {
--- a/pkg/cypher/executor_mutations_relationship_delete.go
+++ b/pkg/cypher/executor_mutations_relationship_delete.go
@@ -79,7 +79,7 @@ func (e *StorageExecutor) tryExecuteBoundRelationshipDelete(
 	}
 	if len(sources) == 0 {
 		result := &ExecuteResult{Columns: []string{}, Rows: [][]interface{}{}, Stats: &QueryStats{}}
-		e.applyDeleteReturnProjection(result, cypher, deleteVar)
+		e.applyDeleteReturnProjection(result, cypher, deleteVar, singleDeleteProjectionInfo(deleteVar, deleteProjectionRelationship))
 		return result, true, nil
 	}
 
@@ -146,7 +146,7 @@ func (e *StorageExecutor) tryExecuteBoundRelationshipDelete(
 		}
 		result.Stats.RelationshipsDeleted = len(edgeIDs)
 	}
-	e.applyDeleteReturnProjection(result, cypher, deleteVar)
+	e.applyDeleteReturnProjection(result, cypher, deleteVar, singleDeleteProjectionInfo(deleteVar, deleteProjectionRelationship))
 	return result, true, nil
 }
 
@@ -298,10 +298,7 @@ func (e *StorageExecutor) boundRelationshipDeleteTargetNode(
 		return nil, false, nil
 	}
 	target, err := store.GetNode(targetID)
-	if err != nil {
-		return nil, false, err
-	}
-	if target == nil {
+	if err != nil || target == nil {
 		return nil, false, nil
 	}
 	return target, true, nil
--- a/pkg/cypher/executor_mutations_relationship_delete.go
+++ b/pkg/cypher/executor_mutations_relationship_delete.go
@@ -156,6 +156,10 @@ func (e *StorageExecutor) collectBoundRelationshipDeleteSources(ctx context.Cont
 	}
 
 	store := e.getStorage(ctx)
+	if txWrapper, ok := store.(*transactionStorageWrapper); !ok || !txWrapper.tx.HasPendingNodeMutations() {
+		return e.collectNodesWithStreaming(ctx, sourcePattern.labels, sourcePattern.properties, sourcePattern.variable, "", -1)
+	}
+
 	var nodes []*storage.Node
 	var err error
 	if len(sourcePattern.labels) > 0 {
--- a/pkg/cypher/executor_mutations_relationship_delete.go
+++ b/pkg/cypher/executor_mutations_relationship_delete.go
@@ -1,0 +1,315 @@
+package cypher
+
+import (
+	"context"
+	"strconv"
+	"strings"
+
+	"github.com/orneryd/nornicdb/pkg/storage"
+)
+
+func (e *StorageExecutor) tryExecuteBoundRelationshipDelete(
+	ctx context.Context,
+	matchSegment string,
+	cypher string,
+	deleteVars string,
+	detach bool,
+) (*ExecuteResult, bool, error) {
+	matchSegment = strings.TrimSpace(matchSegment)
+	_, inTransactionWrapper := e.getStorage(ctx).(*transactionStorageWrapper)
+	if detach || strings.Contains(deleteVars, ",") {
+		return nil, false, nil
+	}
+	deleteVar := strings.TrimSpace(deleteVars)
+	if deleteVar == "" {
+		return nil, false, nil
+	}
+	limit := -1
+	baseMatchSegment := matchSegment
+	if !isSimpleBoundRelationshipDeleteMatchSegment(matchSegment) {
+		limitedSegment, limitedRows, ok := parseBoundRelationshipDeleteWithLimitSegment(matchSegment, deleteVar)
+		if !ok {
+			return nil, false, nil
+		}
+		baseMatchSegment = limitedSegment
+		limit = limitedRows
+	}
+
+	pseudo := baseMatchSegment + " RETURN __delete_probe__"
+	returnIdx := findKeywordIndex(pseudo, "RETURN")
+	if returnIdx <= 0 {
+		return nil, false, nil
+	}
+	whereIdx := lastKeywordIndexBefore(pseudo, "WHERE", returnIdx)
+	clauses := splitMatchClauses(pseudo, whereIdx, returnIdx)
+	if len(clauses) != 2 {
+		return nil, false, nil
+	}
+
+	sourcePattern := e.parseNodePattern(ctx, clauses[0])
+	if sourcePattern.variable == "" {
+		return nil, false, nil
+	}
+	relPatternText := strings.TrimSpace(clauses[1])
+	relMatch := e.parseTraversalPattern(ctx, relPatternText)
+	if relMatch == nil || relMatch.IsChained {
+		return nil, false, nil
+	}
+	if relMatch.StartNode.variable != sourcePattern.variable {
+		return nil, false, nil
+	}
+	if relMatch.Relationship.Variable != deleteVar {
+		return nil, false, nil
+	}
+	if relMatch.Relationship.MinHops != 1 || relMatch.Relationship.MaxHops != 1 {
+		return nil, false, nil
+	}
+	if relMatch.Relationship.Direction != "outgoing" && relMatch.Relationship.Direction != "incoming" && relMatch.Relationship.Direction != "both" {
+		return nil, false, nil
+	}
+
+	whereClause := ""
+	if whereIdx > 0 && whereIdx < returnIdx {
+		whereClause = strings.TrimSpace(pseudo[whereIdx+len("WHERE") : returnIdx])
+	}
+
+	sources, err := e.collectBoundRelationshipDeleteSources(ctx, sourcePattern, inTransactionWrapper)
+	if err != nil {
+		return nil, true, err
+	}
+	if len(sources) == 0 {
+		result := &ExecuteResult{Columns: []string{}, Rows: [][]interface{}{}, Stats: &QueryStats{}}
+		e.applyDeleteReturnProjection(result, cypher, deleteVar)
+		return result, true, nil
+	}
+
+	typeSet := buildRelTypeSet(relMatch.Relationship.Types)
+	store := e.getStorage(ctx)
+	edgeIDs := make([]storage.EdgeID, 0)
+	seen := make(map[storage.EdgeID]struct{})
+	for _, source := range sources {
+		if source == nil {
+			continue
+		}
+		candidates, err := e.boundRelationshipDeleteCandidateEdges(store, source.ID, relMatch.Relationship.Direction)
+		if err != nil {
+			return nil, true, err
+		}
+		for _, edge := range candidates {
+			if edge == nil || !relationshipTypeAllowed(edge.Type, relMatch.Relationship.Types, typeSet) {
+				continue
+			}
+			if len(relMatch.Relationship.Properties) > 0 && !e.edgeMatchesProps(edge, relMatch.Relationship.Properties) {
+				continue
+			}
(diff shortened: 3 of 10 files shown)


orneryd (changes requested):
**Findings**
1. High: the new delete fast path can ignore `WITH/LIMIT/ORDER BY/SKIP` between `MATCH` and `DELETE`.
   PR #230 inserts `tryExecuteBoundRelationshipDelete` before the existing WITH/LIMIT delete hot path, and the helper only checks that `splitMatchClauses` returns two `MATCH` patterns. `parseTraversalPattern` ignores trailing text after the end node, so a shape like:
   ```cypher
   MATCH (a)
   MATCH (a)-[r]->(b)
   WITH r LIMIT 1
   DELETE r
   ```
   can be parsed as a plain bound relationship delete and delete every matching adjacent relationship, not the limited row set. I’d gate this fast path with the same “simple match segment only” restrictions as the streaming delete path, or move it after the WITH/LIMIT handler. Add a regression test for `MATCH ... MATCH ... WITH r LIMIT 1 DELETE r`.

2. High: yes, snapshot isolation is shaky here, but the nuance is important.
   temporal_viewport.go itself delegates to the active storage object. For `MATCH (n:Label)`, a transaction wrapper eventually reaches `BadgerTransaction.GetNodesByLabel`, which uses the read timestamp. For labelless `MATCH (n)`, it reaches badger_transaction.go, which calls `tx.engine.AllNodes()` directly, so that path can see current heads instead of the transaction snapshot.

3. High: the new relationship-delete fast path leans on adjacency reads that are not snapshot-pinned.
   In transactions, `GetOutgoingEdges` / `GetIncomingEdges` route to badger_transaction.go, and those call the engine’s current adjacency indexes rather than visible-at topology APIs. Since the PR runs the new fast path even when `inTransactionWrapper` is true, a delete can discover and delete a relationship committed after the transaction began. Either disable this fast path inside transaction wrappers for now, or make adjacency candidate discovery snapshot-aware before enabling it for writes.

4. Medium: indexed source-node discovery is also not fully SI-safe.
   `collectNodesWithStreaming` can use current schema property indexes. Calling `GetNode` afterward filters out post-begin creations, but it cannot recover false negatives when a node/property index entry existed at the transaction snapshot and was deleted or updated after the transaction began. For a delete fast path, that can silently miss rows. This is related to the same candidate-discovery issue: candidate enumeration needs to be snapshot-aware, not just candidate materialization.

Essentially the fast path needs a stricter eligibility gate and transaction/snapshot behavior is fixed for this. The optimization is directionally good, but it is currently too easy for it to delete more than the Cypher row pipeline selected.

linuxdynasty:
Thanks for the detailed review. I pushed an update on top of current `main` at `c4901451`.

What changed in response:

- The relationship-delete fast path now has a strict segment gate. It only accepts a simple two-`MATCH` delete, or `WITH <rel> LIMIT <positive integer>` between the match and delete. It rejects `ORDER BY`, `SKIP`, `CALL`, `UNWIND`, and `OPTIONAL MATCH` for this path.
- Added a regression for `MATCH ... MATCH ... WITH rel LIMIT 1 DELETE rel`; it deletes one relationship and leaves one behind.
- Fixed the transaction snapshot issue at the storage layer for `BadgerTransaction.AllNodes`, `GetOutgoingEdges`, and `GetIncomingEdges`. Those reads now use MVCC-visible committed state when the transaction has a read timestamp, then merge pending transaction writes.
- Kept the delete fast path enabled in transaction wrappers only after making its reads snapshot-safe. In the transaction-wrapper case, source-node discovery intentionally avoids current property indexes and filters snapshot-visible label/all-node results instead.
- Added a transaction regression where a second relationship is committed after the transaction begins; the delete fast path deletes only the pre-begin relationship and leaves the post-begin relationship intact.

Proof run locally after rebasing and installing the local llama library (`./scripts/build-llama.sh b9835`):

- `go test ./pkg/cypher -run 'TestMatchPatternProperty_(BoundRelationshipDeleteSegmentEligibility|TryBoundRelationshipDeleteDeletesSimpleMatch|TryBoundRelationshipDeleteDeletesUnionAndLimitedMatches|ExecuteDeleteUsesBoundRelationshipDelete|BoundRelationshipDeleteRoutingPredicates|BoundRelationshipDeleteHonorsWithLimit|BoundRelationshipDeleteUsesTransactionSnapshot|BoundRelationshipDeleteUsesAdjacency|BoundRelationshipDeleteReturnsCandidateLookupErrors|TryBoundRelationshipDeletePropagatesCandidateLookupError|BoundRelationshipDeleteReturnsTargetLookupError)' -count=1 -v`\n- `go test ./pkg/storage -run 'TestTransaction_(AllNodesRepeatabilityWithinSnapshot|AdjacencyRepeatabilityWithinSnapshot|LabelScanRepeatabilityWithinSnapshot|EdgeTraversalRemainsSnapshotConsistentAcrossConcurrentDelete)' -count=1 -v`\n- `go test ./pkg/cypher -count=1`\n- `go test ./pkg/storage -count=1`\n- `go test ./pkg/cypher -run '^TestBenchmarkMatchCreateDelete$' -count=1 -v` -> `43341.64 ops/sec`, `0.023 ms/op` on this machine\n- `git diff --check`\n\nOne honest trade-off: transaction-wrapper adjacency now uses the MVCC visible-at edge view instead of the current adjacency index. That is intentional for correctness until a snapshot-aware adjacency index exists; the non-transactional path still uses the existing adjacency index.\n

orneryd (changes requested):
SO I was reviewing this and there is a performance cliff:

with a transaction snapshot, bound relationship delete now reads a snapshot-visible edge set and filters it in memory, which can degrade from node-local work (roughly degree-based) to graph-global work (edge-count based) on large graphs.

The cliff is coming from snapshot tx adjacency reads effectively doing all-edges visible-at plus in-memory filtering by node and direction, so cost trends toward O(E) per expansion instead of O(deg(node)).

To fix the cliff correctly, with backward compatibility preserved:

1. Add native snapshot adjacency APIs to storage Define outgoing/incoming visible-at methods in types.go and wire them through wrappers/delegators (namespaced.go, wal_engine.go, async_engine.go, executor.go). Result: existing engines remain compatible via optional interface + fallback, but the fast path can use exact snapshot semantics.

2. Implement MVCC adjacency index in Badger Build directional snapshot-aware adjacency index structures in Badger MVCC commit/read paths (badger_mvcc.go, badger_edges.go, badger_transaction.go). Result: snapshot adjacency reads become node-local (degree-based), eliminating graph-wide scans.
3. Replace transaction fallback scans with API usage Update transaction read helpers to call the new visible-at adjacency APIs directly instead of visible-at all-type scans plus in-memory filtering (pkg/storage/badger_transaction_snapshot_reads.go). Result: no more O(E) behavior on tx snapshot adjacency.
4. Route Cypher delete fast path to snapshot adjacency directly Keep current eligibility semantics, but force bound relationship delete candidate discovery to use directional visible-at adjacency in tx contexts (pkg/cypher/executor_mutations_relationship_delete.go). Result: fast path remains correct and fast under SI/MVCC.
5. Expand regression and benchmark coverage for final behavior Add/extend correctness tests and high-edge-count performance benchmarks for tx snapshot adjacency and bound relationship delete (transaction_snapshot_anomalies_test.go, match_pattern_property_index_test.go). Result: prevents correctness regressions and proves cliff removal quantitatively.
6. Keep compatibility contract explicit Do not remove existing methods or behavior; new APIs are additive. Engines not implementing snapshot adjacency continue to work through existing paths, but Badger gets the full optimized behavior. Result: backward compatible by design, with the correct fix as the default on primary engine.

Then we can re-evalute any paths that should also be updated. If you want I can make the APIs available and then you can update your PR to make use of them,

linuxdynasty:
Agreed, let's do it that way. Your cliff read is right — snapshot tx adjacency doing visible-at all-edges plus in-memory filter is O(E) per expansion, not O(deg(node)); that's the same pole we hit downstream.

Split works for me:

- You take 1–2: the native snapshot adjacency APIs in `types.go` / wrappers, and the MVCC directional adjacency index in Badger (`badger_mvcc.go`, `badger_edges.go`, `badger_transaction.go`). You're closer to the MVCC commit/read paths and I don't want to guess at the index layout.
- I'll take 3–5 on this PR: route the tx snapshot reads (`badger_transaction_snapshot_reads.go`) and the bound-relationship-delete candidate discovery (`executor_mutations_relationship_delete.go`) to the new visible-at adjacency APIs, and add the high-edge-count regression + benchmark coverage to prove the cliff is gone (degree-local, not edge-count).
- 6 we both hold: additive only, no removed methods, optional interface + fallback for engines without snapshot adjacency.

I'll hold #230 as-is — not merging until your APIs land, since the current fast path can hit exactly the O(E) tx-snapshot cliff you flagged. Ping me when the interface is up, or push it to a branch I can build against, and I'll wire the consumer side + benchmarks on top.

Appreciate you digging into this properly instead of rubber-stamping the fast path.

orneryd:
just merged #253 into main, you will want to rebase and then wire up the changes for the steps 3-5. Thanks for you patience too I really want to make the API correct and stable especially when it comes to deletes. 😓

linuxdynasty:
Handling it now ;)

linuxdynasty:
Wired the consumer side onto the snapshot adjacency APIs now that they're on main. Pushed as `8985c046`.

3 + 4 collapsed into one change. `getCommittedAdjacentEdgesLocked` was resolving snapshot adjacency with `GetEdgesByTypeVisibleAt("", readTS)` and filtering by endpoint in memory — the O(E) tx-snapshot cliff. Swapped it to the directional `GetOutgoingEdgesVisibleAt`/`GetIncomingEdgesVisibleAt(nodeID, readTS)`, so it's O(deg(node)) now. The bound-relationship-delete fast path didn't need its own edit: candidate discovery goes `store.GetOutgoingEdges` -> `transactionStorageWrapper` -> `BadgerTransaction.GetOutgoingEdges` -> the same helper, so it goes node-local in tx contexts for free.

5, proof. New Badger-backed regression in `badger_transaction_snapshot_reads_test.go` (node-local read returns only the bound edge among 200 unrelated, stable across a concurrent commit, fresh tx sees the new one), plus a high-edge-count benchmark with the bound node pinned at degree 1:

```
                before        after
edges=100       85.5 us/op    53.8 us/op
edges=1000       969 us/op    52.5 us/op
edges=5000      5.40 ms/op    50.5 us/op   (~107x)
allocs         2.3k->85.6k    667 flat
```

Linear in total edge count before, flat after. That's the cliff gone, measured degree-local not edge-count.

Verify:
- `go test ./pkg/storage ./pkg/cypher -count=1` -> ok
- `go test ./pkg/storage -bench BenchmarkTransaction_SnapshotAdjacency -benchtime=200x` -> numbers above
- existing `AdjacencyRepeatabilityWithinSnapshot` + `EdgeTraversalRemainsSnapshotConsistentAcrossConcurrentDelete` still green

Rebased on current main so this sits on top of the merged adjacency index. Thanks for landing 1–2 — the directional API dropped straight in.

orneryd:
I found another issue i had to push up to main in the helper implementations, i aligned the behaviors to be the same across the package:


aligned the dangling-endpoint behavior by extracting the endpoint lookup into a shared traversal helper and keeping the existing semantics explicit: if an adjacency entry points at a node that no longer loads, that candidate is skipped instead of failing the whole query. The change is in pkg/cypher/traversal.go, and the regression test is in pkg/cypher/chained_traversal_test.go.

That gives the future bound-relationship delete helper a single place to reuse the current traversal contract rather than re-implementing its own endpoint handling. Validation passed with focused cypher tests, including [TestTraverseFromNode](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) and the delete-helper suite.


you'll need to rebase again, sorry 🫶

linuxdynasty:
Pushed `54618531` to cover the remaining transaction-source side of the bound relationship delete path.

What changed:
- `collectBoundRelationshipDeleteSources` now reuses the existing indexed/streaming pattern-inline source lookup in transaction-wrapper mode when the transaction has no pending node mutations.
- If the transaction has pending node mutations, it keeps the old `GetNodesByLabel` fallback so read-your-writes behavior is preserved.
- Added regression coverage for both sides:
  - indexed transaction-mode source lookup does not leak a `GetNodesByLabel` scan
  - pending transaction source nodes are still visible
- Added a benchmark for the transaction-mode bound-delete source lookup as repository label population grows.

Why this matters:
The previous follow-up made snapshot adjacency node-local, but the source node lookup could still take the transaction-mode label-scan branch before adjacency expansion. For shapes like:

```cypher
MATCH (source_repo:Repository {id: $repo_id})
MATCH (source_repo)-[rel:CORRELATES_DEPLOYABLE_UNIT]->(:Repository)
WHERE rel.evidence_source = $evidence_source
DELETE rel
```

the source anchor should be resolved by the declared property index when there are no pending tx node writes.

Proof run locally on darwin/arm64:

```text
go test ./pkg/cypher ./pkg/storage -run 'TestMatchPatternProperty_BoundRelationshipDelete(TransactionSourceUsesIndex|TransactionSourceKeepsPendingNodes|UsesAdjacency|SegmentEligibility|RoutingPredicates|HonorsWithLimit|UsesTransactionSnapshot|ReturnsCandidateLookupErrors|ReturnsTargetLookupError)|TestTransaction_SnapshotAdjacency' -count=1 -v
PASS: pkg/cypher, pkg/storage

go test ./pkg/cypher ./pkg/storage -count=1
ok github.com/orneryd/nornicdb/pkg/cypher 28.003s
ok github.com/orneryd/nornicdb/pkg/storage 43.327s
```

Benchmark before/after for the source lookup benchmark (`BenchmarkMatchPatternProperty_BoundRelationshipDeleteTransactionSourceLookup`, `-benchmem`):

```text
Before, repositories=100:  ~49.6-50.5 us/op,   95,764 B/op,  1,346 allocs/op
After,  repositories=100:  ~0.76-0.83 us/op,    1,208 B/op,     21 allocs/op

Before, repositories=1000: ~507-612 us/op,    933,010 B/op, 13,052 allocs/op
After,  repositories=1000: ~0.67-0.95 us/op,    1,208 B/op,     21 allocs/op

Before, repositories=5000: ~3.10-3.98 ms/op, 4,792,352 B/op, 65,062 allocs/op
After,  repositories=5000: ~1.09-1.53 us/op,    1,208 B/op,     21 allocs/op
```

linuxdynasty:
Rebased onto current `main` (`16c99722`) and pushed `6bfaad33`.

What changed after the rebase:

- Updated the bound relationship delete fast path for the new typed delete-return projection helper, so `DELETE rel RETURN count(rel)` counts relationship deletes correctly on this path.
- Aligned target endpoint lookup with the traversal behavior you just put on `main`: dangling/missing endpoint candidates are skipped instead of failing the whole delete.
- Added a focused gate test file for the fast-path predicates and endpoint helper: strict simple-segment rejection, `WITH rel LIMIT n` parsing/rejection cases, relationship type filtering, direction handling, and dangling endpoint skip behavior.

Verification on the rebased head:

```text
go test ./pkg/cypher ./pkg/storage -run 'TestMatchPatternProperty_BoundRelationshipDelete(TransactionSourceUsesIndex|TransactionSourceKeepsPendingNodes|UsesAdjacency|SegmentEligibility|RoutingPredicates|HonorsWithLimit|UsesTransactionSnapshot|ReturnsCandidateLookupErrors|SkipsTargetLookupError)|TestTransaction_SnapshotAdjacency|TestIsSimpleBoundRelationshipDeleteMatchSegment|TestParseBoundRelationshipDeleteWithLimitSegment|TestBoundRelationshipDeleteTargetNode|TestRelationshipTypeAllowed|TestTraverseFromNode' -count=1 -v
PASS: pkg/cypher, pkg/storage

go test ./pkg/cypher ./pkg/storage -count=1
ok github.com/orneryd/nornicdb/pkg/cypher 37.309s
ok github.com/orneryd/nornicdb/pkg/storage 53.107s

git diff --check
PASS
```

Re-run benchmark checks on this rebased head:

```text
BenchmarkTransaction_SnapshotAdjacency
edges=100:  67.0-75.8 us/op, ~31.7-32.3 KB/op, 667-668 allocs/op
edges=1000: 68.8-83.0 us/op, ~31.7-31.9 KB/op, 667 allocs/op
edges=5000: 78.8-98.4 us/op, ~31.7-31.9 KB/op, 667 allocs/op

BenchmarkMatchPatternProperty_BoundRelationshipDeleteTransactionSourceLookup
repositories=100:  1027-1146 ns/op, 1208 B/op, 21 allocs/op
repositories=1000: 1129-1156 ns/op, 1208 B/op, 21 allocs/op
repositories=5000: 1063-1261 ns/op, 1208 B/op, 21 allocs/op
```

So both pieces still look flat with graph/label population after the rebase: snapshot adjacency is degree-local, and transaction-mode source lookup stays index-based when there are no pending node writes.

pkg/storage/badger_transaction.go
```diff
@@ -1382,7 +1382,7 @@ func (tx *BadgerTransaction) GetOutgoingEdges(nodeID NodeID) ([]*Edge, error) {
 		return nil, err
 	}
 
-	committed, err := tx.engine.GetOutgoingEdges(nodeID)
+	committed, err := tx.getCommittedAdjacentEdgesLocked(nodeID, "outgoing")
```
orneryd: since this is along a fast path, we should be using enums here instead of raw magic strings for comparison.

orneryd: something like this

```

type EdgeDirection uint8

const (
    Outgoing EdgeDirection = iota
    Incoming
)

func (tx *BadgerTransaction) getCommittedAdjacentEdgesLocked(
    nodeID NodeID,
    direction EdgeDirection,
) ([]*Edge, error) {

    if tx.readTS.IsZero() {
        switch direction {
        case Outgoing:
            return tx.engine.GetOutgoingEdges(nodeID)
        case Incoming:
            return tx.engine.GetIncomingEdges(nodeID)
        default:
            return nil, ErrInvalidData
        }
    }

    switch direction {
    case Outgoing:
        return tx.engine.GetOutgoingEdgesVisibleAt(nodeID, tx.readTS)
    case Incoming:
        return tx.engine.GetIncomingEdgesVisibleAt(nodeID, tx.readTS)
    default:
        return nil, ErrInvalidData
    }
}

```

linuxdynasty: Done in `982eddc6`. Added `type EdgeDirection uint8` (`Outgoing`/`Incoming`) and moved `getCommittedAdjacentEdgesLocked` plus both call sites (`GetOutgoingEdges`/`GetIncomingEdges`) off the magic strings, so the fast path compares the integer tag now. Unknown value still returns `ErrInvalidData`, and I kept the enum next to the function with godoc. `go vet` + `go test ./pkg/storage -count=1` green.

orneryd (approved):

PR #261, opened 2026-07-13
Fix UNIQUE commit lock false sharing

linuxdynasty:
## What changed

Replace the fixed 256-stripe UNIQUE commit-lock table with an active exact-key registry.

Transactions that touch the same `(label, property, value)` still serialize across constraint validation, Badger commit, and unique-cache publication. Disjoint write batches no longer serialize because unrelated values happened to hash to the same stripe.

The registry:

- reserves references for holders and waiters before blocking, preventing entry deletion/ABA;
- assigns every active exact key a stable total order, preventing AB-BA deadlocks;
- removes entries after the last holder or waiter;
- skips non-reflexive values such as NaN, which cannot conflict under equality and cannot be safely deleted from a Go map.

This addresses the concurrent graph-write timeout shape retained in [Eshu #5122](https://github.com/eshu-hq/eshu/issues/5122) without reducing worker or writer concurrency.

## Same-machine proof

| Proof | Shipped stripes | Exact-key registry |
| --- | ---: | ---: |
| 8 writers x 100 disjoint keys admitted together | 1/8 | 8/8 |
| Lock-window benchmark median | ~10.1 ms/op | ~1.41 ms/op |
| Real Badger commit benchmark median | ~10.16 ms/op | ~6.89 ms/op |
| Same-UID concurrent commits | one winner, retryable loser | one winner, retryable loser |
| Final Function UID exactness | — | 800 rows, 800 distinct UIDs, cache mappings exact |

The focused lock benchmark is about 7.2x faster. The realistic cache-complete Badger commit benchmark is about 32% lower latency on the same machine.

A local Eshu end-to-end run kept graph concurrency enabled (8 graph writes in flight and 16 Nornic entity workers):

- 77,690 entities plus 4,579 statements in the larger repository scope;
- canonical graph write: 14.54 seconds;
- 830 Function chunks, typically 0.10–0.22 seconds each;
- queue terminal: 33 succeeded, 0 dead;
- graph readback: 121,576 nodes, 271,964 edges;
- Function exactness: 87,130 rows, 87,130 distinct non-null UIDs;
- no graph-write timeout, claimed-concurrent-commit, dead-letter, restart, or OOM signal.

That is local representative proof, not a claim that the full remote corpus is complete.

## Verification

```text
go test -tags 'noui nolocalllm' ./pkg/storage -run '^Test(BadgerUniqueConstraintCommit|UniqueConstraintCommitLocks)' -count=3
go test -race -tags 'noui nolocalllm' ./pkg/storage -run '^Test(BadgerUniqueConstraintCommit|UniqueConstraintCommitLocks)' -count=3
go test -tags 'noui nolocalllm' ./pkg/storage -count=1
golangci-lint run --build-tags noui --new-from-rev=8603da03a921112dbc9d426d0cd3aaa15b5a8713 ./pkg/storage
```

The repository-wide headless suite passes when excluding `pkg/localllm`, `pkg/heimdall`, and `TestLoadPluginsFromDir`. The unchanged baseline fails those surfaces because the local Darwin Llama archive is absent; the same baseline plugin test was run to confirm that limitation.

## Review

Cold review found no P0, P1, or P2 issue. Tests cover disjoint batches, overlapping lock order, holder/waiter lifecycle, registry cleanup, NaN cleanup, real Badger commits, same-UID conflict/retry behavior, and exact unique-cache readback.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Large disjoint UNIQUE-constrained write batches no longer serialize on
+  hash-lock collisions.**
+  Replaced the fixed 256-stripe commit-lock table with an active,
+  reference-counted exact-value registry. Transactions touching the same
+  `(label, property, value)` still serialize through validation, Badger commit,
+  and unique-cache publication, while disjoint batches commit concurrently.
+  Registry-assigned ordering prevents AB-BA deadlocks, entries expire after the
+  last holder or waiter, and non-reflexive values such as NaN cannot leak lock
+  entries. Added real Badger commit, same-value retry, race, cleanup, and
+  8-writer/100-key performance coverage.
 - **Multi-MATCH relationship variable bound in a later clause was silently
   dropped.**
   `MATCH (s) WHERE s.uid IN $u MATCH (s)-[rel]->() WHERE
--- a/pkg/storage/badger_transaction.go
+++ b/pkg/storage/badger_transaction.go
@@ -2727,23 +2727,22 @@ func (tx *BadgerTransaction) checkTemporalConstraint(node *Node, c Constraint) e
 }
 
 // acquireUniqueConstraintCommitLocks collects the unique constraints touched
-// by this transaction's pending nodes and acquires bounded
-// per-(label, property, value) mutex stripes on the transaction's pinned
-// namespace's schema. Returns a release function that unlocks in reverse
-// order; safe to defer.
+// by this transaction's pending nodes and acquires exact
+// per-(label, property, value) mutexes on the transaction's pinned namespace's
+// schema. Returns a release function that unlocks in reverse order; safe to
+// defer.
 //
 // Locks are keyed by value, not only by constraint: a transaction adding
-// nodes with uids "X" and "Y" hashes those values to bounded lock stripes; a
-// peer transaction with uids "X" and "Z" serializes against "X" and commits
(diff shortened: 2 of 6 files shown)


linuxdynasty:
Remote Eshu validation on the exact PR commit is complete for the original commit-conflict path.

Run shape:

- 896 repositories, 894 valid heads, 2 empty-history repositories
- graph-write in-flight 8
- projection 8, reducer 16, shared 4, source-local projector 2
- no worker or concurrency reduction

Initial bootstrap/readback result:

| Signal | Result |
|---|---:|
| UNIQUE violations | 0 |
| concurrent-commit claims | 0 |
| graph-write timeout / context deadline | 0 |
| panic / fatal | 0 |
| service restarts / OOM | 0 / 0 |
| Repository nodes | 896 |
| Function nodes | 510,367 |
| Function UID total-minus-distinct | 0 |
| null Function UID | 0 |

This supports the narrow conclusion from the local proof: exact-key serialization prevents conflicting same-UID commits while preserving disjoint-key concurrency. It does not by itself make the whole Eshu corpus terminal; a separate Eshu generation-liveness defect later reopened 790 obsolete source-local rows and caused canonical-retract timeouts.

CI distinction: Build passes. The Test and Coverage job's code/tests passed on each run, but the final Coveralls upload returned HTTP 504 three times. I am not retrying the same external upload failure again.

linuxdynasty:
CI rerun is now fully green on the unchanged proof commit `1492458852588c884c32f70d27ea2ee07086769c`: Build, Test and Coverage, and Coveralls all passed. The exact-key/local and 896-repository evidence in the prior comment is unchanged. This is ready for the requested maintainer review.

PR #263, opened 2026-07-17
fix(cypher): fail loud on unsupported multi-clause query shapes

linuxdynasty:
## Summary

Multi-clause read queries that the string-slicing WITH interpreters cannot faithfully evaluate previously returned **corrupt results silently** — the worst failure mode for a graph database, since a caller cannot distinguish a wrong answer from a right one. This adds a **fail-loud guard** that converts those specific unsupported shapes into a clear, actionable error (mirroring the existing `unsupported clause after CALL {}` contract), so the failure is visible and debuggable instead of silent.

## The bug (before)

For a set of multi-clause read shapes, `executeMatchWithClause` / `executeMatchWithOptionalMatch` / `executeCompoundMatchOptionalMatch` returned:
- raw expression text as a column value (e.g. a value of `"DISTINCT a.name"`, `"length(path)"`, `"min(length(path))"`),
- a bool/nil where a real value was expected,
- skipped `DISTINCT` deduplication, or
- dropped all rows.

## The fix (after)

New `unsupportedMultiClauseShape` (`pkg/cypher/failloud_guard.go`), invoked at the three multi-clause handler entry points, detects these shapes and returns `unsupported multi-clause query: <detail>` with an actionable single-clause alternative. **Scoped to multi-clause reads only** — single-clause `RETURN DISTINCT` / `min()`/`max()`, working `WITH`/`OPTIONAL MATCH`/aggregate/`collect()` projections, and `EXPLAIN`/`PROFILE` are untouched.

## Before / after — accuracy (per symptom)

All 7 symptom shapes reproduced as failing tests, then made to error:

| shape | before | after |
|---|---|---|
| multi-clause `RETURN DISTINCT <expr>` | literal `"DISTINCT <expr>"`, dedup skipped | clear error |
| `length(path)` after another clause | literal / `0` | clear error |
| `min`/`max`/aggregate in multi-clause projection | literal text | clear error |
| `WITH … OPTIONAL MATCH` | dropped all rows | clear error |
| comma-separated multi-pattern `MATCH` | dropped all rows | clear error |
| pattern comprehension in projection | `false` | clear error |
| zero-length `*0..N` hop-0 row | literal text | clear error |

## Before / after — performance

Benchmark on the guarded path (uncached full multi-clause `WITH` query): baseline ~92.9 µs/op / 1605 allocs vs. with-guard ~93.6 µs/op / 1613 allocs → **+0.76% / +8 allocs**, within run-to-run noise and far under the 5% budget. The guard runs only on cache-miss executions (isolated detector cost ~497 ns/op).

## Validation (AGENTS.md bar)

- **Full `pkg/cypher` suite passes with ZERO tests flipped** (`go test -tags nolocalllm ./pkg/cypher/`). The one shape predicted to flip (`TestCodeGraphContextRepoVisualizationQuery`) does **not** — its `*0..` under `OPTIONAL MATCH` returns a correct typed subgraph, so Rule 7 is scoped to the primary `matchSection` to avoid that false positive.
- **New-code coverage:** `unsupportedMultiClauseShape` 97.8%; helpers 100%. Package total unchanged at 90.2%.
- **`-race`:** clean.
- **`gofmt`:** clean on touched files. `go vet`: clean.
- **CHANGELOG.md** updated.

Files: `pkg/cypher/failloud_guard.go` (new), `pkg/cypher/failloud_multiclause_test.go` (new), `pkg/cypher/match_with.go` (+16), `pkg/cypher/clauses.go` (+8), `CHANGELOG.md`.

Downstream context: discovered while hardening Eshu's query layer against these exact shapes; Eshu is separately rewriting its affected reads to single-clause forms, so this guard is defense-in-depth (silent-corruption → visible error), not a hard dependency.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,30 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   aggregate per-row mutation counters (nodes/relationships created/deleted,
   properties set, labels added) into Bolt result summaries so downstream
   clients observe accurate counters.
+- **Unsupported multi-clause Cypher shapes now fail loud instead of returning
+  corrupt results.**
+  The legacy string-slicing WITH interpreters
+  (`executeMatchWithClause` / `executeMatchWithOptionalMatch` /
+  `executeCompoundMatchOptionalMatch`) silently produced corrupt output for a
+  set of multi-clause read shapes they cannot evaluate — raw expression text as
+  a column value, a boolean/nil where a value was expected, skipped `DISTINCT`
+  deduplication, or an empty result — with no error, so a caller could not tell
+  a wrong answer from a right one. A new guard (`unsupportedMultiClauseShape` in
+  `pkg/cypher/failloud_guard.go`) detects exactly these shapes at handler entry
+  and returns a clear, actionable error mirroring the existing
+  "unsupported clause after CALL {}" contract. Rejected shapes: `RETURN DISTINCT`
+  after `WITH`/`OPTIONAL MATCH`; `length(<path>)` projected across an
+  `OPTIONAL MATCH`; `min()`/`max()`/`avg()` over a node pattern in a multi-clause
+  `RETURN`; aggregation in `WITH` followed by `OPTIONAL MATCH`; comma-separated
+  multi-pattern `MATCH` combined with `WITH`; pattern comprehensions in a
+  multi-clause projection; and zero-length variable-length patterns (`*0..`) in
+  the primary `MATCH` consumed by `WITH`. Single-clause queries and supported
+  multi-clause shapes (bare `WITH` + `OPTIONAL MATCH`, `count()`/`sum()`/
+  `collect()` projections, `min()`/`length()` over relationship paths, `*0..`
+  inside `OPTIONAL MATCH`) are unaffected. Added `failloud_multiclause_test.go`
+  with failing-then-green coverage for all seven shapes plus edge cases; guard
+  overhead measured at ~0.76% on an uncached multi-clause `WITH` query
+  (93.6 µs vs 92.9 µs/op), well within the no-regression budget.
 
 ## [v1.1.11] - 7/9/2026
 
--- a/pkg/cypher/clauses.go
+++ b/pkg/cypher/clauses.go
@@ -3558,6 +3558,14 @@ type optionalRelResult struct {
 // This implements left outer join semantics for relationship traversals with aggregation support
 func (e *StorageExecutor) executeCompoundMatchOptionalMatch(ctx context.Context, cypher string) (*ExecuteResult, error) {
 	originalCypher := cypher
+
+	// Fail loud on multi-clause shapes this string interpreter cannot evaluate
+	// correctly (e.g. RETURN DISTINCT after OPTIONAL MATCH), rather than
+	// emitting corrupt rows. See unsupportedMultiClauseShape for detail.
+	if detail, unsupported := e.unsupportedMultiClauseShape(cypher); unsupported {
+		return nil, fmt.Errorf("unsupported multi-clause query: %s", detail)
+	}
+
 	params := getParamsFromContext(ctx)
 	// Substitute parameters AFTER routing to avoid keyword detection issues
 	if params != nil {
--- a/pkg/cypher/failloud_guard.go
+++ b/pkg/cypher/failloud_guard.go
@@ -1,0 +1,288 @@
+package cypher
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+// Fail-loud guard for the legacy multi-clause WITH interpreters.
+//
+// A subset of multi-clause Cypher queries is handled by the string-slicing
+// interpreters in match_with.go (executeMatchWithClause /
+// executeMatchWithOptionalMatch and their delegates). For a handful of query
+// shapes those interpreters cannot faithfully evaluate the projection, and
+// historically they returned CORRUPT results silently instead of erroring:
+//
+//   - raw expression text as a column value (e.g. the literal string
+//     "DISTINCT a.name" or "b.name"),
+//   - a boolean/nil where a real value was expected,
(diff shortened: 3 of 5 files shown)


orneryd:
I have major concerns here with trying to reject valid query shapes instead of fixing issues or adding support.

orneryd:
Findings:

1. This PR deliberately rejects valid Neo4j query shapes, so it moves NornicDB away from drop-in compatibility rather than toward it. The new guard hard-errors on `RETURN DISTINCT` after `WITH`/`OPTIONAL MATCH`, aggregation before `OPTIONAL MATCH`, and `*0..` traversals in the primary match at [pkg/cypher/failloud_guard.go], [pkg/cypher/failloud_guard.go], and [pkg/cypher/failloud_guard.go], and those errors are wired into the live handlers at [pkg/cypher/match_with.go]) and [pkg/cypher/clauses.go] I checked the local neo4j source: its parser grammar accepts these constructs, and its planner tests explicitly cover `RETURN DISTINCT` after `OPTIONAL MATCH` and `*0..` carried across `WITH`. If we accept this PR, NornicDB will return “unsupported multi-clause query” for queries that Neo4j executes.

2. High: the blacklist runs before the more general pipeline delegation, so it can reject queries outside the narrow legacy interpreter shape it claims to protect. In [pkg/cypher/match_with.go] the guard fires before the code reaches the pipeline handoff at [pkg/cypher/match_with.go]. That means a query like `MATCH (a) WITH a MATCH (b) RETURN DISTINCT a.name, b.name` will be blocked by Rule 1 solely because the final projection starts with `DISTINCT`, even though this is a general pipeline shape, not just the fragile `MATCH ... WITH ... RETURN` string-slicing case. So this PR does not just “fail loud instead of corrupting”; it also freezes whole valid classes of queries out of any better executor behind the router.

3. Medium: the detector is not parser-aware, so it will still miss some corrupting inputs and falsely reject some valid ones. The strongest false positive is Rule 7: `zeroLengthVarRe` is a raw regex over the entire pre-`WITH` text at [pkg/cypher/failloud_guard.go] and [pkg/cypher/failloud_guard.go], so a valid query like `MATCH (a {note:'*0..'}) WITH a RETURN a` would be rejected even though the `*0..` is only data. The strongest false negative is Rule 1: `leadingKeyword` at [pkg/cypher/failloud_guard.go] only trims whitespace, so `RETURN /* comment */ DISTINCT ...` bypasses the guard and can still fall into the corrupt interpreter path. That is especially concerning because this package already has comment/string-aware scanners in [pkg/cypher/helpers.go] and [pkg/cypher/keyword_scan.go]; the new guard does not use the same level of parsing hygiene.

4. Medium: Rule 3’s “node-only vs relationship-pattern” split is wrong for bare-arrow patterns, so it will reject queries the PR itself intends to preserve. The discriminator only looks for bracketed relationship syntax at [pkg/cypher/failloud_guard.go], but both Neo4j and this codebase accept bracketless relationships like `(a)-->(b)`. The new tests even acknowledge bare-arrow pattern comprehension as valid syntax at [pkg/cypher/failloud_multiclause_test.go]. So `MATCH p=(a)-->(b) WITH p RETURN min(length(p))` would be classified as “node pattern + min” and rejected by Rule 3, even though the test suite claims the relationship-path case should remain supported at [pkg/cypher/failloud_multiclause_test.go].

orneryd (changes requested):
We should support known query shapes, this pulls us further away from the goal

linuxdynasty:
> We should support known query shapes, this pulls us further away from the goal

Makes total sense! I will re-review and see how to move from here.

linuxdynasty:
You're right, and I'd rather change course than patch the guard.

The premise was wrong. Fail-loud only makes sense for shapes we genuinely can't support. Using it on shapes Neo4j executes pushes us away from drop-in compatibility, which is the whole point. Rejecting a valid query isn't really better than corrupting it. Either way the caller's working Neo4j query breaks.

Your findings hold up. Naming them so it's clear I read them and I'm not hand-waving:

Rejecting valid shapes is the actual defect. RETURN DISTINCT after WITH/OPTIONAL MATCH, aggregation before OPTIONAL MATCH, and *0.. carried across WITH are all valid Neo4j, and the grammar plus planner tests confirm it. Erroring on them is a compatibility regression, not a safety win.
The guard fires before the pipeline handoff, so it locks valid pipeline queries like MATCH (a) WITH a MATCH (b) RETURN DISTINCT a.name, b.name out of any better executor behind the router. That's the worst part, because it forecloses the correct fix.
The detector isn't parser-aware. Rule 7's raw regex false-positives on a *0.. that's just string data, and Rule 1's whitespace-only leadingKeyword false-negatives on RETURN /* comment */ DISTINCT. And it reinvents scanning the package already has in helpers.go / keyword_scan.go.
Rule 3 misclassifies bare-arrow patterns. MATCH p=(a)-->(b) WITH p RETURN min(length(p)) is a relationship path, not "node + min", so it rejects a shape this PR claims to keep supported.
One thing that changes the math: the corruption vector this was guarding on the rel-existence path is already root-caused and fixed in [#264](https://github.com/eshu-hq/eshu/issues/264) (bracket-less rel-existence predicates, OPTIONAL MATCH rel-var resolution, non-DETACH DELETE guard), and that merged. So the "silent corruption to visible error" argument for this guard is mostly moot now.

Plan:

Closing this one. A narrowed guard still rejects shapes we should support, so there's nothing here worth keeping as-is.
Opening a follow-up that adds real support: route these multi-clause shapes through the pipeline executor instead of the legacy string-slicing interpreter, so RETURN DISTINCT after WITH/OPTIONAL MATCH, aggregate-before-OPTIONAL MATCH, and carried *0.. actually execute and return correct results. Failing-test-first against the exact shapes above, using the existing comment/string-aware scanners instead of new regexes.
On the Eshu side I'll keep the affected reads on single-clause forms in the meantime so nothing depends on this landing.
I'll re-scope and come back with the support PR. Thanks for checking this against the actual Neo4j grammar and planner behavior. That's the bar it should've been held to from the start.

PR #264, opened 2026-07-18
fix(cypher): correctly evaluate relationship-existence predicates and guard non-detach DELETE

linuxdynasty:
## Problem

Three related correctness bugs in the Cypher WHERE/DELETE evaluation cause relationship-existence predicates to be evaluated wrong and let a non-DETACH `DELETE` silently destroy relationships. They were found while an Eshu graph-cleanup job (eshu-hq/eshu#5147) got the wrong answer on every node.

1. **Relationship-existence predicates in `WHERE`/`COUNT`/`EXISTS` are mis-evaluated for common pattern shapes.** The gate that decides whether a `WHERE` fragment is a relationship pattern required a bracket *and* an explicit arrow (`-[` plus `]->`/`<-`). So the bracket-less forms `(n)--()`, `(n)-->()`, `(n)<--()`, and the bracketed-undirected `(n)-[r]-()` fell through to a fallback that treats the fragment as a plain expression and returns true. The result: `WHERE NOT (n)--()` matches **nothing** (the negation of always-true), `WHERE (n)--()` matches **everything**, and `COUNT { (n)--() } = 0` is a tautology that matches every node regardless of its edges. A cleanup that deletes "nodes with no relationships" using these predicates therefore either deletes nothing or deletes connected nodes.

2. **A relationship variable bound by an embedded `OPTIONAL MATCH` resolves to `nil`.** The mutation paths (`executeDelete`/`executeSet`/`executeRemove`) build an internal `MATCH ... OPTIONAL MATCH (a)-[r:T]->(b) RETURN ...` probe and run it through the low-level `executeMatch`. `parseTraversalPatternStateMachine` locates the relationship bracket by scanning forward from the first node group's closing paren instead of searching for `-[`, so the literal `OPTIONAL MATCH (n)` text between the two patterns corrupts the substring handed to `parseRelationshipPattern`, and the relationship variable/type/properties are silently dropped — `r` comes back `nil` even when the `OPTIONAL MATCH` genuinely matched. (A second, independent gap surfaced here: `executeRemove` had no `*storage.Edge` support at all and duplicated rows when binding two or more node variables.)

3. **A non-DETACH `DELETE` of a connected node silently cascade-deletes its relationships.** `DeleteNode` removes a node and all its edges, and the generic delete path never checked for residual relationships, so `DELETE n` behaved like `DETACH DELETE n` and destroyed edges the caller never named — and didn't even count them.

## Fix

1. `containsRelExistencePattern` / `bareRelDirection` (new `rel_existence_pattern.go`) classify bracket-less and bracketed-undirected patterns and hand them to the existing relationship evaluator, which already resolves direction correctly. Wired into both `evaluateWhere` and `evaluateWhereOnPath`, and into the `COUNT`/`EXISTS` subquery paths.
2. `executeMatch` routes a query whose `matchPart` contains a relationship pattern **and** an embedded `OPTIONAL MATCH` through `executeCompoundMatchOptionalMatch` — the same handler the top-level dispatcher already uses — so the relationship variable resolves to the real edge (`nil` only when the `OPTIONAL MATCH` genuinely didn't match). `executeRemove`/`applyRemoveToMatchedRows` gain `*storage.Edge` support and one-row-per-match handling, mirroring the proven `SET` path.
3. The generic delete path now validates before it mutates: for a non-DETACH `DELETE` it errors (openCypher / Neo4j semantics — `Cannot delete node, because it still has relationships`) unless those relationships are themselves being deleted in the same statement. This is a **behavior change**, documented in `CHANGELOG.md` with a one-line migration note (use `DETACH DELETE`).

## Proof

- Failing-first regression tests for every case (table-driven): bare/directed/undirected rel-existence in `WHERE`, `COUNT`/`EXISTS` bare bodies, `OPTIONAL MATCH` rel-var resolution via the raw probe path (real-match and genuine-no-match), `SET`/`REMOVE` on an `OPTIONAL MATCH` relationship, and non-DETACH `DELETE` of a connected node erroring while `DETACH DELETE` succeeds.
- The Neo4j compatibility suite stays 100% green, including the previously-masked `Delete_chunks_with_OPTIONAL_MATCH` case, now restored to a plain `DELETE r, chunk` as the green proof that the root cause is fixed, not worked around.
- Full `pkg/cypher` + `pkg/bolt` suites green; race-clean on the changed surface; total coverage held at 90.2%.

Local build note: this box can't link the llama libs, so the suites were run with `-tags nolocalllm`; CI hydrates the libs and runs the full build + coverage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,53 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **WHERE relationship-existence predicates now recognize bracket-less and
+  undirected patterns, and bare `COUNT`/`EXISTS` subquery bodies.**
+  `WHERE (n)--()`, `WHERE (n)-->()`, `WHERE (n)<--()`, and the bracketed
+  undirected form `WHERE (n)-[r]-()` previously fell through to the
+  relationship-pattern gate's default branch, which treats an unrecognized
+  expression as `true` -- so `WHERE NOT (n)--()` matched nothing and
+  `WHERE (n)--()` was always true, regardless of the graph's actual shape.
+  Separately, `COUNT { (n)--() }` and `EXISTS { (n)--() }` required their
+  subquery body to start with `MATCH`, so a bare (unprefixed) body always
+  returned `0` / `false`. Both gates now recognize the full range of
+  bracket-less and undirected existence patterns.
+- **Non-DETACH `DELETE` of a node that still has relationships now errors
+  instead of silently cascade-deleting its edges (behavior change).**
+  `MATCH (n) DELETE n` previously called the storage engine's
+  `DeleteNode` unconditionally, which cascade-deletes every adjacent edge --
+  so a plain `DELETE` on a connected node quietly removed its relationships
+  too, diverging from openCypher/Neo4j semantics. `DELETE` now validates that
+  every node in the deletion plan has no relationships left outside the
+  statement's own edge deletions *before* mutating anything (so a multi-row
+  `DELETE` is judged as a whole, not partially applied), and errors with the
+  same "still has relationships" wording Neo4j uses. Deleting a node together
+  with its own edge in one statement (`DELETE a, r`) still works, since the
+  edge being removed no longer counts as residual. Existing callers relying on
+  the old silent cascade must switch to `DETACH DELETE`.
+- **An `OPTIONAL MATCH`-bound relationship variable resolved to `nil` inside
+  `DELETE`/`SET`/`REMOVE`'s internal match probes, even when the `OPTIONAL
+  MATCH` genuinely matched.** `executeDelete`, `executeSet`, and
+  `executeRemove` each build an internal
+  `MATCH ... OPTIONAL MATCH (a)-[r:TYPE]->(b) RETURN <vars>` probe and execute
+  it via the low-level match path instead of the dispatcher that normally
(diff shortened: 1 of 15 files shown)


linuxdynasty:
### Proven end-to-end through Eshu against a pr264 image

Built a `cpu-bge` image from this branch's head and ran Eshu's own Bolt driver (DB `nornic`, autocommit `session.Run`) against it, with the same discriminating fixture used to characterize the bug: an `orphan` File (no edge) and `connected`-[:CONTAINS]->`peer`.

| predicate | pinned v1.1.11 | **this PR (pr264)** |
| --- | --- | --- |
| `MATCH (n) WHERE NOT (n)--() DELETE n` | orphan survives (no-op) | **orphan deleted, connected + peer preserved** ✅ |
| `MATCH (n) WHERE COUNT { (n)--() } = 0 DELETE n` | deletes everything (over-delete) | **orphan deleted, connected + peer preserved** ✅ |

So the relationship-existence predicates evaluate correctly on this branch, confirmed not just in `pkg/cypher` unit tests but through the real Bolt path a downstream (Eshu) uses. This is the root-cause fix behind eshu-hq/eshu#5147.

(Separately: the `ExecuteWrite`/managed-transaction `UNWIND ... MATCH (a)-[r]->(b) DELETE r` under-apply — a different bug — still reproduces on this branch, i.e. this PR does not address it; tracked as eshu-hq/eshu#4902.)

PR #265, opened 2026-07-23
fix(cypher): support every traversal-seeded OPTIONAL MATCH + aggregate shape by mirroring Neo4j

linuxdynasty:
## What's broken

Do a `MATCH` that binds a relationship, then a trailing `OPTIONAL MATCH` with no `WITH` between them, and the `RETURN` comes back wrong. `type(rel)` comes back as the literal string `"type(rel)"`, the `rel` variable never binds, chained `OPTIONAL MATCH` clauses get dropped, and aggregates don't evaluate. So a query like "give me this class's INHERITS edges plus the file each one lives in" silently returns garbage for the computed columns.

My first pass at this failed loud on the shapes it didn't handle, which was the wrong call. This version supports every valid shape by doing what Neo4j does. I pulled the Neo4j source (`community/cypher`) and copied the real semantics instead of guessing.

## Behavior, before and after

| Shape | Before | After |
|---|---|---|
| `type(rel)`, `coalesce(...)`, `head(labels(...))` after a trailing OPTIONAL MATCH | returned as literal text (`"type(rel)"`) | evaluated |
| bare `rel` / `rel.prop` | literal text (`"rel"`) | bound to the real edge |
| chained 2nd+ OPTIONAL MATCH | silently swallowed | applied |
| disconnected OPTIONAL MATCH — `MATCH (a)-[r]->(b) OPTIONAL MATCH (x:X) RETURN a, x` | rejected (fail-loud) | supported: left-outer join, null-fill on no match |
| rebound rel var on parallel edges — `MATCH (a)-[r:T]->(b) OPTIONAL MATCH (a)-[r]->(b) RETURN count(*)` | fanned out over parallel edges, overwrote `r` | identity constraint: one row per originally bound edge (`count(*)=2`, no overwrite) |
| aggregates — `stDev`, `STDEVP`, `count(tf)+1`, `coalesce(sum(w),0)` | errored / narrowed set | full executor registry, case-insensitive |

## How it works now

The Neo4j operators I mirrored (`community/cypher/`):

- `OptionalPipe` / `ApplyPipe` / `OptionalExpandAllPipe` — for each left row, run the inner pattern; on no match keep the row and null the new variables. This is the general Apply+Optional path that handles disconnected, multi-hop, and pre-bound-rel-var shapes.
- A rel var already in scope is an identity constraint, not a fresh expansion — same as `planSingleProjectEndpoints` in `expandSolverStep.scala`, and the same invariant NornicDB already asserts at `multi_match_relationship_binding_bug_test.go:364-375`.
- `EagerAggregationPipe` + the front-end `isolateAggregation` rewrite for mixed aggregate expressions, and `StdevFunction` (Welford) for `stdev`/`stdevp`.

The only thing that still errors is `count()` with no argument, which Neo4j rejects at parse time too. The old fail-loud-multiclause commit is dropped; this branch is rebased onto `main`.

## Performance

Kept the pre-compilation pass from the earlier tuning. Wall-clock had too much machine noise to trust, so these are the deterministic alloc counts (3-way interleaved, cache-off, precompiled binaries):

| Shape | corrupt `main` | pre-support | support |
|---|---|---|---|
| SingleSeedChained allocs/op | 2241 / 154.9KB | 2352 / 166.3KB | 2353 / 166.3KB |
| FanOutProjection allocs/op | 6459 / 549.4KB | 6175 / 588.2KB | 6175 / 588.2KB (below corrupt baseline) |

The support rework adds nothing measurable over the pre-compilation pipeline, and fan-out still allocates less than the original corrupt path.

## Tests

The four corruption cases go red without the fix and green with it, plus a control, the fix-behavior tests, and named regression tests for each shape the maintainer flagged — `TestSupport_DisconnectedOptionalBareVarReturn`, `TestSupport_ReboundRelVariableIsIdentityConstraint`, `TestSupport_StdevMixedCaseAggregate`. Full `pkg/cypher` and `pkg/bolt` are green, coverage is 90.3% (the four pipeline files 96.6-100%), and I rebuilt the binary and re-checked it live over Bolt.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   row per matched row regardless of how many node/relationship variables are
   in scope. The same fix applies to a chained `SET ... REMOVE ...` clause in a
   single statement.
+- **Traversal-seeded `OPTIONAL MATCH` projections are now evaluated instead of
+  echoed as literal expression text.**
+  A read query whose primary `MATCH` contains a relationship pattern followed
+  by one or more trailing `OPTIONAL MATCH` clauses (no intervening `WITH`)
+  previously resolved its `RETURN` items with a string resolver that only
+  understood `var.prop` and bare variables. Every other expression came back
+  as its own source text: `RETURN type(rel)` returned the literal string
+  `"type(rel)"`, `coalesce(...)`/`labels(...)`/`head(...)` returned their
+  source text, aggregates like `count(f)` returned the string `"count(f)"`,
+  the primary MATCH's relationship variable was dropped entirely (so
+  `rel.weight` returned `"rel.weight"`), chained second-level `OPTIONAL MATCH`
+  clauses were silently swallowed (their bindings projected as literal text),
+  and per-clause `WHERE` predicates were ignored. The path now executes the
+  seed `MATCH` with relationship variables bound, left-outer joins every
+  chained `OPTIONAL MATCH` clause in either direction (seeding from whichever
+  endpoint is bound), evaluates projections through the real expression
+  evaluator (with a fast path for plain `var.prop`/bare-variable items),
+  routes aggregate projections through implicit-grouping aggregation with
+  Cypher null semantics, and applies `ORDER BY`/`SKIP`/`LIMIT`. An
+  `OPTIONAL MATCH` clause that references no previously bound variable now
+  fails loud instead of returning an empty result.
 - **Bolt explicit transactions now bind top-level `UNWIND` rows before
   routing to mutation handlers.**
   `session.ExecuteWrite` queries shaped as `UNWIND ... MATCH ... DELETE`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,9 +74,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   endpoint is bound), evaluates projections through the real expression
   evaluator (with a fast path for plain `var.prop`/bare-variable items),
   routes aggregate projections through implicit-grouping aggregation with
-  Cypher null semantics, and applies `ORDER BY`/`SKIP`/`LIMIT`. An
-  `OPTIONAL MATCH` clause that references no previously bound variable now
-  fails loud instead of returning an empty result.
+  Cypher null semantics, and applies `ORDER BY`/`SKIP`/`LIMIT`. The clause
+  semantics mirror Neo4j's runtime operators: a connected single-hop clause
+  behaves like `OptionalExpandAll` (null seeds propagate null bindings), and
+  every other shape — a disconnected pattern sharing no variable with earlier
+  clauses, a single-node pattern, a multi-hop chain, or a pre-bound
+  relationship variable — is evaluated with `Apply` + `Optional` semantics:
+  matches extend the row, and a row with no match is preserved once with
+  newly-introduced variables bound to null. No valid shape is rejected.
+  Aggregation follows Neo4j's model end to end: implicit grouping by the
+  non-aggregate items, identity values over empty ungrouped input, `stdev`/
+  `stdevp` per Neo4j's `StdevFunction`, and `RETURN` items that contain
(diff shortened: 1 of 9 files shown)


orneryd:
flagging a few issues.

- High: the new traversal-specific OPTIONAL MATCH engine regresses Neo4j/Cypher semantics by failing when a trailing OPTIONAL MATCH clause has no previously bound variable, instead of treating it as a legal disconnected optional pattern. The touched dispatch point is clauses.go:3627. I checked the local Neo4j source as requested: it has planner coverage for OPTIONAL MATCH as a standalone legal clause and shell diagnostics that explicitly describe using OPTIONAL MATCH for disconnected patterns, which only makes sense if those patterns are valid and produce optional cartesian behavior. This PR’s new “fail loud” branch would reject valid shapes like MATCH (a)-[r]->(b) OPTIONAL MATCH (x:X) RETURN a, x on exactly the traversal path it is trying to fix.

- High: the new join logic appears to rebind an already-bound relationship variable when both endpoints are already bound, which is a correctness bug on graphs with parallel edges. The existing codebase already treats “same relationship variable must still mean the same edge” as an invariant, with explicit regression coverage at multi_match_relationship_binding_bug_test.go:364 and multi_match_relationship_binding_bug_test.go:375. In the new OPTIONAL MATCH engine, the described behavior only checks whether the already-bound target node matches; it does not preserve an existing rel binding. So a query shaped like MATCH (a)-[r]->(b) OPTIONAL MATCH (a)-[r]->(b) RETURN count(*) would fan out across every matching edge between a and b and silently overwrite r, instead of keeping one row per originally bound edge.

- Medium: the new aggregate path narrows supported aggregate functions for this one query shape and will start erroring on aggregates the rest of the executor already recognizes. The current aggregate registry includes stdev and stdevp at clauses.go:424, and evaluator coverage exists at functions_test.go:1961. But the PR description says the new parser only supports count, sum, avg, min, max, and collect. That means traversal-seeded MATCH ... OPTIONAL MATCH ... RETURN stDev(...) would now fail loudly while the rest of NornicDB still treats it as a supported Cypher aggregate. That is a behavior regression, not just an implementation detail.

linuxdynasty:
Thanks for the detailed review. These were all against the earlier revision — the branch you saw still had the fail-loud commit and my original rejection paths. I've since rebased onto main (dropping the fail-loud commit) and reworked it to support the shapes by mirroring the Neo4j runtime. Walking your three points against the current head, each pinned with a named test you can run:

1. Disconnected OPTIONAL MATCH. The fail-loud branch is gone, and so is my "references no variable bound by an earlier clause" error. MATCH (a)-[r]->(b) OPTIONAL MATCH (x:X) RETURN a, x runs fine now — the disconnected clause goes through the Apply+Optional path, so a match binds and a no-match keeps the left row with x null (OptionalPipe/ApplyPipe). Test: TestSupport_DisconnectedOptionalBareVarReturn.

2. Rebound rel var on parallel edges. You were right about the code you looked at (the old both-bound filter only checked the target node). On the current head a pre-bound rel var routes through the general path where r is shared and candidates have to agree on edge identity, so it's an identity constraint, not a fresh expansion. On two parallel :T edges, MATCH (a)-[r:T]->(b) OPTIONAL MATCH (a)-[r]->(b) RETURN count(*) returns 2 — one row per originally bound edge — and the parallel edge doesn't overwrite r. That matches the invariant at multi_match_relationship_binding_bug_test.go:364-375 and ProjectEndpoints in the planner (a rel var already in scope gets endpoint projection, not a fresh expandSolverStep). Test: TestSupport_ReboundRelVariableIsIdentityConstraint.

3. stdev/stdevp. The traversal aggregate set matches the executor registry (collect/count/sum/avg/min/max/stdev/stdevp), so nothing the rest of NornicDB recognizes errors here. RETURN stDev(r.n), STDEVP(r.n) evaluates case-insensitively (0.7071… / 0.5 on {1,2}). The only remaining error is count() with no argument, which Neo4j rejects at parse time too. Test: TestSupport_StdevMixedCaseAggregate.

Full pkg/cypher and pkg/bolt are green, coverage's at 90.3%. Force-pushed.

orneryd (approved):

PR #353, opened 2026-09-12
feat(search): add native filtered search continuation

### LINKED ISSUE #346: Feature request: native continuation for filtered vector and hybrid search
## Use case

We are evaluating NornicDB for a browsable media library, combining metadata filters, lexical search, and externally generated text/visual embeddings. Search is also a way to browse the collection: users need access beyond the leading relevance shortlist.

For example, a filtered collection might contain 20,000 eligible assets, displayed 50 at a time. The leading text candidates may be reranked with Voyage, but later results must remain accessible without reranking the entire collection. Multiple passages or frames may belong to one displayed asset.

## Feature request

Would you consider native continuation/pagination for filtered vector and hybrid retrieval, or recommend an existing supported query pattern that already provides it efficiently?

The current `SearchOptions` interface appears to expose a top-K limit and bounded candidate retrieval, but no continuation cursor. We would like a supported way to request subsequent bounded batches without repeatedly increasing the initial top-K or carrying a growing list of already-returned IDs in application code.

Useful behavior would be:

- Stable score/ID ordering across batches for unchanged data and query settings, with no repeated or skipped members of the declared result population.
- The same metadata filters applied throughout continuation.
- An explicit distinction between exhausting the selected result population and reaching an approximate-search candidate budget. A short ANN response should not by itself claim that every eligible result has been visited.
- A documented policy for continuation when indexed data changes; invalidating a cursor and requiring a fresh query would be acceptable for our use case. We are not asking for a long-lived immutable snapshot.
- Guidance on grouping passage/frame hits into distinct parent assets across batches, if supported.

Our catalogue's eligible-asset counts are separate from the ANN shortlist size. For collection browsing, we need an explicit route to continue through the remaining eligible assets after the leading ranked pool, including those not selected by ANN. This could be an exact ordered query/continuation path rather than an approximate-search guarantee. We do not expect ANN to guarantee exhaustive recall, or a reranker to rank the whole collection globally.

## Why it would help

This could keep retrieval and continuation in the database instead of requiring application-owned result caches, repeated larger searches, or exclusion-list management. It would be useful for browsable catalogues as well as search interfaces that only display a few top hits.

This is separate from #344, which reports correctness and execution-time problems in existing filtered cosine query forms. Here the question is the supported continuation contract and whether it can be provided natively. If an existing efficient Cypher or retrieval pattern already covers this, an example would be very helpful.

mm0nst3r:
Closes #346.

Add native Go and Cypher continuation for filtered search and eligible-collection browsing. `db.retrieve.page` freezes a bounded result population and serves scope-bound cursors without repeatedly increasing top-K or growing exclusion lists. Callers choose ranked results, all eligible IDs, or a ranked prefix followed by an explicitly unranked remainder. Mutations invalidate affected retained populations; unrelated database changes preserve their scope.

This PR is based directly on current upstream `main`, with no dependency on #355 or #356. It contains only continuation and its required integration. Optional explanatory/diagnostic search metadata is preserved generically within the existing cursor resource bounds, without provider-specific types or retained source nodes/vectors.

Verification covers standalone Linux search/continuation race tests and actual HTTP/Bolt use with 101 original Library records: complete 15-page enumeration, cursor replay, mutation invalidation, persistence/restart and unchanged source text, across both BM25 engines and synchronous/asynchronous storage. The combined build also passed actual Voyage retrieval with exact passages and stable provider reports on first, next, replayed and grouped pages; standalone continuation emits no provider-specific extras. Actual provider checks also verified per-request truncation/failure policy against opposite server defaults, stable reports on continuation/replay, and cursor rejection when either policy changes. No full Library capacity or cross-instance cursor support is claimed. Broader repository runs encountered upstream test/build-profile and resource failures; this is not a blanket whole-suite PASS.

Contracts and resource limits: `docs/user-guides/search-continuation.md`.


orneryd:
@mm0nst3r I am fixing some of the bolt continuation semantics first. There's a neo4j compatibility gap I am closing there for qid-based continuation from the same connection. I realize this is a stronger idea but I want to explore the performance implications of this at the same time.

mm0nst3r:
Superseded by #375, which carries continuation forward through the shared QID API. Closing this proposal in favour of the upstream implementation.

Issue #348, opened 2026-09-11
Feature request: managed Voyage AI embeddings with context-4 auto-chunking and multimodal support


mm0nst3r:
## Proposal

Please consider native Voyage AI support in NornicDB's managed embedding pipeline, covering both contextualized text embeddings and multimodal embeddings. These fit one provider integration, although the two input paths could be delivered separately if that suits the implementation.

This complements the Voyage reranker request #345; it concerns generating and managing embeddings, rather than only storing vectors supplied by an application.

## Why this could be a particularly strong fit for NornicDB

From our perspective, Voyage is substantially ahead of the alternatives we have considered on retrieval quality, while being very inexpensive. The attraction is not simply another provider checkbox: combining Voyage with Nornic's graph, vector, lexical retrieval and reranking could make Nornic a much more complete, broadly useful, near drop-in RAG backend.

**The central appeal is specifically `voyage-context-4`: send the full document text, receive usable chunks and contextualized embeddings, and let Nornic store, index and retrieve them. No separately deployed chunker, chunking model or application-side chunking pipeline is needed for that path.** At the published embedding price below, that is a compelling alternative to operating additional infrastructure just to prepare chunks.

For applications that already have document text, this could remove the chunking stage of a Docling-based or similar RAG preparation pipeline. [Docling also performs parsing, layout extraction and OCR](https://docling-project.github.io/docling/); this proposal replaces the separate chunking responsibility, not those document-conversion functions. The context model can also reduce the need for additional LLM calls to enrich individual chunks with surrounding context, making the combination useful across document types and domains.

MongoDB's [context-4 announcement](https://www.mongodb.com/company/blog/product-release-announcements/voyage-context-4-stop-worrying-about-chunking-with-our-best-performing-model) describes this design, reports evaluation across 39 datasets spanning eight domains, and lists pricing of $0.12 per million tokens. Our enthusiasm about quality/value is our assessment, not a claim that one model wins every possible workload.

## 1. Contextualized text and automatic chunking

Support `/v1/contextualizedembeddings`, with `voyage-context-4` as the concrete use case:

- Submit a full document with `enable_auto_chunking=true`, retaining configurable chunk size and overlap when wanted.
- Preserve the returned chunk text, order, embeddings and association with the original document, so retrieval can return the actual supporting passage.
- Generate query embeddings with `input_type=query`; document generation uses `input_type=document`.
- Expose output dimensions, including 2048, and keep the model configurable.
[...]

Issue #347, opened 2026-09-11
Feature request: optional Snowball stemming as an extension of Unicode BM25 analysis


mm0nst3r:
## Proposal

Would you consider optional Snowball stemming as a lightweight extension of NornicDB's existing Unicode-aware BM25 text analysis?

This looks like an almost drop-in integration at the linguistic-algorithm level: reuse established Snowball stemmers after Unicode normalization and tokenization, rather than developing language rules in NornicDB. Snowball already supports generating Go code and supplies language algorithms and reference test data. No external service, language model, or dictionary-development project is needed.

There is still integration work around configuration, using the same analyzer for indexing and querying, and persisted-index settings; we are not claiming that the complete feature is literally a one-line change.

## Fit with existing Unicode support

The current tokenizer in `pkg/search/fulltext_index.go` applies NFKC normalization, Unicode case folding, and Unicode-aware token splitting. An opt-in Snowball token-processing step would extend that foundation to matching inflected word forms.

This also fits the existing [language-agnostic BM25 plan](https://github.com/orneryd/NornicDB/blob/main/docs/plans/bm25-language-agnostic-retrieval-quality-plan.md), which keeps the default language neutral while allowing optional language-specific analyzers. Stemming is an additional search behavior, not a correction to Unicode normalization itself.

## Concrete use case: Russian

Our immediate use case is Russian-language media metadata and transcripts. For example, the established Russian stemmer maps `вагон`, `вагона`, and `вагоны` to the same stem. The current official Russian algorithm also handles `ё` to `е` folding, useful for spelling variants such as `ёлка` / `елка`.

The rules and expected outputs already exist upstream; implementing this would not require designing Russian morphology or speaking Russian. The same integration could support other Snowball languages.

## Suggested scope

- Keep the existing language-neutral analyzer as the default; make the Snowball language an explicit configuration choice.
- Apply the selected analyzer consistently to indexed text and queries.
- Preserve original text for display and reranking; stemming only affects lexical index terms.
- Include the analyzer identity/settings in persisted-index compatibility, using the existing rebuild mechanism when index-time settings change.
- Use Snowball's reference vocabulary/output fixtures plus a few indexing/query examples to verify the integration. Automatic language detection and new stopword policies need not be part of this request.

Native support would remove our need to maintain a separate application-side normalized-text representation just to feed BM25.

## Existing implementation and test resources

- [Snowball project and supported code-generation languages](https://snowballstem.org/)
[...]

Issue #345, opened 2026-09-11
Feature request: native Voyage AI reranker provider


mm0nst3r:
# Feature request: native Voyage AI reranker provider

Please add Voyage AI as a supported reranking provider, so applications can use NornicDB retrieval followed by Voyage reranking without maintaining an HTTP response translator or a separate reranking integration.

Voyage's official API is `POST https://api.voyageai.com/v1/rerank`:
https://docs.voyageai.com/reference/reranker-api

It accepts `query`, `documents`, `model`, `top_k`, `return_documents` and `truncation`, with bearer authentication. Results use a top-level `data` array containing original document `index` and `relevance_score`, together with model/usage information. Models should be configurable rather than hardcoded; current stable examples are `rerank-2.5` and `rerank-2.5-lite`.

The inspected generic HTTP reranker in `pkg/search/rerank.go` currently sends `top_n` and parses `results`, `scores` or `rankings`. That is not the Voyage request/response contract. Merely changing the endpoint does not establish native Voyage support.

Requested integration:

- Use the existing reranker abstraction and configuration, with a Voyage provider implementation and securely configured API credential.
- Map `data[].index` back to the original candidate identities, preserving scores, ordering and the requested returned count.
- Respect Voyage's input limits and caller truncation choice; do not silently discard text to satisfy a limit.
- Preserve useful provider status/request/usage/error information without exposing credentials or submitted content in logs. Authentication, rate-limit, timeout or malformed-response failures must not be presented as successful reranking; respect an explicit caller failure policy.
- Include mocked contract tests using synthetic text and document the configuration and supported request/response behavior.

This is text reranking: candidates may originate from dense retrieval, lexical/BM25 retrieval or their combination. Voyage receives the candidate text, not sparse vectors or image embeddings. Independent visual similarity can remain outside this reranking step.

No credentials or application data are needed in this issue. An illustrative synthetic response is:

```json
{"data":[{"index":2,"relevance_score":0.8},{"index":0,"relevance_score":0.5}],"model":"rerank-2.5","usage":{"total_tokens":100}}
```

mm0nst3r:
Implemented in #355. We built and tested the contributions together using actual Voyage calls and our original Library text/image data, including full supporting passages, rerank results, pagination and restart preservation.

The implementation work started before we saw your design comment on #347.

mm0nst3r:
[...]

PR #354, opened 2026-09-12
fix(cypher): invalidate cached reads after public and replicated node changes

mm0nst3r:
Public or replicated node deletion could succeed while a warmed Cypher query continued to return the old node/count. Those writes do not necessarily pass through the executor that owns the cached result.

Include the existing database-scoped node mutation revision in cache identity, preserving the revision captured before execution through result insertion. Forward that optional revision capability through the size-tracking and replication wrappers. The storage engine remains the revision owner; no second cache or counter is introduced.

Tests cover direct and queued-update deletion, flush/reopen, unchanged cache hits, database isolation, a read overlapping deletion, and actual StorageAdapter publication into a warmed query. The regression fails before correction and passes afterward, including race detection. The combined actual Voyage primary/standby run confirms that a successful primary delete is subsequently absent from the warmed standby query.

---
Part 6 of 8 in the tested contribution series. Depends on https://github.com/orneryd/NornicDB/pull/353. [Focused change for this part](https://github.com/mm0nst3r/NornicDB/compare/0469c3379d55c55032418697a48c0830b646216b...f04e6bd4d65567f9edf327fafc57ef2344f78f8b). GitHub's cumulative diff also shows preceding parts until they are merged; the focused comparison isolates this contribution.

The [complete locally verified source](https://github.com/mm0nst3r/NornicDB/tree/944eeb93b20d4604a8a65d9d4615465192d7f838) retains all contributions. The optional Snowball part is last so Voyage can be accepted independently of the stemming design. No production deployment has been performed.


orneryd:
Heya, thanks for pointing this out. this PR correctly identifies stale Cypher cache results when direct or replicated mutations bypass the executor. Its focused fix is node-specific and depends on the stacked continuation work in #353.

Our standalone fix removes that dependency and broadens invalidation to all graph mutations: nodes, relationships, async/in-flight writes, replication, bulk prefix deletion, namespaces, wrappers, and Fabric constituents. It also keeps label-count metadata coherent and publishes revisions only after mutation cleanup completes.

that work is fixed in 0a7830f1.

As for new features I would rather see a proposal with charts showing the changes before and after the intended changes before applying any new feature-specific work.

orneryd:
FYI you will need to rebase your branches for the windows-specific fixes to merge in

mm0nst3r:
The diagrams and architecture proposals are now in the relevant issue discussions:

- [#345 — Voyage reranking](https://github.com/orneryd/NornicDB/issues/345#issuecomment-5648597286)
- [#346 — Search continuation](https://github.com/orneryd/NornicDB/issues/346#issuecomment-5648597411)
- [#347 — Revised language-agnostic stemming plugins](https://github.com/orneryd/NornicDB/issues/347#issuecomment-5648619287)
- [#348 — Voyage embeddings](https://github.com/orneryd/NornicDB/issues/348#issuecomment-5648597603)

PR #355, opened 2026-09-12
feat(voyage): add managed embeddings and native reranking

### LINKED ISSUE #345: Feature request: native Voyage AI reranker provider

### LINKED ISSUE #348: Feature request: managed Voyage AI embeddings with context-4 auto-chunking and multimodal support

mm0nst3r:
Closes #345. Closes #348.

Add native Voyage reranking and managed contextualized/multimodal embeddings through the existing provider, worker, search and storage owners. Provider-returned text stays associated with its source, with explicit failure/truncation policies and outcome reports.

Rebased onto #375 (`search-continuation`, upstream head `77a2798`), which is this PR's base branch; it will retarget `main` once #375 merges. It contains no continuation endpoints or cursor state of its own. Query chunks use the upstream common search helper; native reranking runs once after fusion and follows #375's candidate-budget contract (the rerank pool is declared as a bounded boundary, never as retrieval exhaustion).

Voyage text is returned as `supporting_passages` across HTTP, Cypher, MCP and native gRPC (gRPC `SearchHit` field 11), distinct from #375's grouped-child `passages`, and is carried on grouped children too. The native provider records complete candidate coverage before minimum-score filtering. Generic result/response metadata accessors let consumers retain that evidence without depending on Voyage implementation details. These changes are in this PR; no separate integration patch is required.

Verification of the rebased four-PR source:
- Original 101 Library records over HTTP and Bolt: both BM25 engines, neutral/Russian/reopen transitions, updates/deletion/recreation, original-text preservation, and complete 15-page enumeration with replay.
- Actual Voyage calls using the original transcript and JPEG: contextualized and multimodal generation, vector-matched retrieval and supporting text, 101-candidate reranking, settings and lifecycle controls, restart, and the gRPC database model override.
- Linux CPU server build and focused native/transport/search suites. Provider score-filter and paginated/grouped-result checks use a simulated provider; real-provider pagination has not been exercised.

The configured rerank candidate pool remains bounded (default 100, maximum 1,000). No production deployment or whole-repository test-pass claim is made.

Provider and lifecycle contracts: `docs/user-guides/voyage.md`.

Post-review corrections (b7070c70):
- Supporting passages survive the upstream light node projection used by search enrichment (contextualized nodes lacking vectors are re-read in full before the currentness check); previously valid passages were dropped on direct storage engines.
- Continuation cursor byte accounting now includes retained supporting-passage text, so the configured registry and build limits apply to it.
- Stale tests were rewritten against the current worker (claim-based duplicate prevention, conditional embedding publication); the MCP discover test waits on the database-level search readiness barrier.
- Verified on the combined 355+356-over-375 tree: package suites (voyage, embed, config, resultstream, gRPC, search, nornicdb, replication, MCP, Cypher, Bolt, server) and the retained 101-record HTTP/Bolt dataset (both BM25 engines, stemming transitions, full 15-page `id`/`ranked_then_id` enumeration).

Post-review corrections (07782d0):
- Query-embedding failures now fail open on Cypher `db.retrieve` and MCP `discover` as they already did on HTTP and gRPC: the search degrades to BM25 and reports it (`search_method`/`fallback_triggered`). Previously a native provider error failed those two transports closed. The caller's explicit `failClosed` option on `db.retrieve` still fails closed.
- MCP `discover` results gain `fallback_triggered`, mirroring the other transports; a degraded search was previously indistinguishable from a plain keyword search. The dropped fallback *reason* is a shared-search matter reported separately in #381.
- Rebased onto the current `search-continuation` head `77a2798` (no conflicts); package suites (MCP, Cypher, gRPC, Voyage, embed, search, nornicdb) pass on the rebased tree.


orneryd (changes requested):
I see stuff about continuation in this PR which sould also be separate to be reviewable

PR #356, opened 2026-09-12
feat(search): add optional Russian and English Snowball stemming

### LINKED ISSUE #347: Feature request: optional Snowball stemming as an extension of Unicode BM25 analysis

mm0nst3r:
Closes #347.

Add optional Russian and English Snowball stemming after the existing Unicode tokenizer, consistently for indexing, updates, deletion and queries in both BM25 engines. Original text and literal phrase matching remain unchanged. Unset/none preserves neutral analysis. Analyzer identity in snapshots and build settings makes incompatible indexes rebuild through the existing owner.

Rebased onto #375 (`search-continuation`, upstream head `77a2798`), which is this PR's base branch; it will retarget `main` once #375 merges. It is independent of #355. All three feature commits reapplied without patch changes; the diff contains only stemming, its dependency, configuration, tests and documentation. It preserves the process-wide stemming proposal and does not implement the later plugin RFC.

Current verification includes focused stemming/analyzer tests and the four-PR Linux CPU build. HTTP/Bolt checks reused 101 original Library records with both BM25 engines, neutral/Russian/same-language reopen/neutral transitions, inflected queries, updates/deletion/recreation and exact original-text preservation. Complete continuation also returned all 101 records in 15 pages with replay. Actual Voyage checks on the combined source ran with Russian stemming enabled.

No production deployment, fresh full BEIR run or whole-repository test-pass claim is made. Configuration and contracts: `docs/user-guides/fulltext-stemming.md`.


orneryd (changes requested):
I see stuff about continuation in this PR which sould also be separate to be reviewable

PR #361, opened 2026-09-13
fix(cypher): order complete index ties before pagination

### LINKED ISSUE #360: Indexed ORDER BY truncates tied groups before applying secondary sort keys
Indexed `ORDER BY` can select a page using only the first sort key, then apply the remaining keys to an already truncated set. This returns the wrong rows/order when the indexed value is shared by multiple nodes.

Reproduced on a clean CPU/headless Linux build of upstream `145ed415607b10207797f136e90227dc1d1d051f`, over Bolt with the official Neo4j Python driver. No additional PR stack is required.

### Reproduction

In a disposable database, create a property index on `IndexedOrderRepro.title`, then insert a group of nodes sharing that title, with the secondary sort value in reverse insertion order:

```python
from neo4j import GraphDatabase

with GraphDatabase.driver("bolt://localhost:7687", auth=None) as driver:
    with driver.session(database="nornic") as session:
        session.run("CREATE INDEX repro_title FOR (n:IndexedOrderRepro) ON (n.title)").consume()
        rows = [dict(title="same title", rank=720-i, payload=720-i, active=i % 3 == 0)
                for i in range(720)]
        session.run("UNWIND $rows AS row CREATE (n:IndexedOrderRepro) SET n = row",
                    rows=rows).consume()
        actual = [row["payload"] for row in session.run("""
            MATCH (n:IndexedOrderRepro) WHERE n.active = true
            RETURN n.payload AS payload
            ORDER BY n.title ASC, n.rank ASC SKIP 0 LIMIT 24
        """)]
        assert actual == list(range(3, 73, 3)), actual
```

Expected: the first 24 qualifying nodes under the **complete** ordering. Actual: candidates selected before secondary ordering; the assertion fails. The same issue occurs in the `WHERE n.title IS NOT NULL` index path and on later pages. A tied group can exceed the current fixed over-fetch window, so increasing that window is not sufficient.

### Cause

`tryCollectNodesFromPropertyIndexOrderLimit` and `tryCollectNodesFromPropertyIndexNotNullOrderLimit` stop appending nodes at `limit`, making their subsequent `len(nodes) > limit` secondary top-K selection unreachable. The property-index top-K operation can itself truncate inside a tied primary-key group. Sorting only that truncated set cannot recover the proper page.

The fix must finish the relevant primary-key tie, evaluate filtering and all remaining sort keys, and only then select the requested window. It should retain a bounded node window rather than loading all candidate node payloads.

### Proposed fix

[PR #361](https://github.com/orneryd/NornicDB/pull/361) includes the native fix, regression tests, verification results and performance tradeoff.

mm0nst3r:
Indexed node pagination currently truncates contenders by the first indexed property before considering secondary sort keys. A tied title group with reverse secondary ranks can therefore return the wrong page. This change finishes each relevant primary-key tie and applies filtering and all ordering keys before selecting the requested window.

The filtered and `IS NOT NULL` paths share one bounded heap collector. The storage visitor releases index locks before fetching nodes. A general query that exhausts the non-null index before filling its window declines the optimization so normal MATCH evaluation can include qualifying rows with a missing sort property. Explicit `IS NOT NULL` queries retain their index path. Existing traversal tests now distinguish that partial-window fallback from a complete indexed result.

Based directly on upstream `145ed415607b10207797f136e90227dc1d1d051f`; independent of the other open PRs. No application-specific code, configuration or data is included.

### Verification

- GitHub [CI run 34764649523](https://github.com/orneryd/NornicDB/actions/runs/34764649523) passed both Build and Test and Coverage, including the test-and-coverage generation step.

- The issue's unchanged official-driver reproducer fails on clean upstream and passes with the fix.
- Native 720-node regression exceeds the old over-fetch window and checks both secondary directions, filtering, `IS NOT NULL`, and multiple node-page offsets.
- Regression preserves qualifying missing-property rows on first and later pages after partial index exhaustion.
- Large-tie traversal tests cover start/end seed ordering without SKIP, plus the existing affected traversal tests.
- Focused Cypher/storage tests pass with `-race -tags noui,nolocalllm`.
- Unlabelled indexed `IS NOT NULL` queries preserve bounded node reads for `LIMIT`.
- Local Bolt checks cover correct ordered pages and index presence after restart for a 72-row duplicated-title cohort over three pages; 18 application query-contract comparisons pass. Private datasets are not part of this contribution.

The separate local Linux headless run was not fully green: it encountered tag-profile compile errors in heimdall/localllm (also reproduced on clean upstream), missing libllama during plugin test builds, a CRLF/LF snapshot difference, a root-sensitive WAL permission assertion, Cypher and storage suite timeouts/kills, a replication timeout assertion and a killed server test process. Only the tag-profile compile failures have been reproduced on clean upstream.

Windows golangci-lint could not complete type checking with the headless tags because existing localllm implementations redeclare types/functions; lint success is not claimed.

This addresses indexed node-window selection and the traversal seed consumers of that selection. An additional nonzero-SKIP traversal/projection ordering probe fails identically on clean upstream and this candidate, outside the indexed seed path; this PR does not claim to fix that separate behavior.

### Cost

With query caching disabled, a 720-node tied-group page measured about 1.25–1.29 ms and 859 KB/op after the fix, versus 0.127–0.147 ms and 123 KB/op for the incorrect baseline. The correct result examines contenders the old cutoff discarded. Retained node payloads are bounded by the requested window; total allocation is not, because complete group IDs and examined nodes still allocate. Partial general queries may use a full scan to include null-sort rows.

Fixes #360.


orneryd:
thanks for this it exposed a couple other issues and i'll be updating this PR before i merge it in im also running some performance profiling on these hot paths to make sure allocations are as close to 0 as possible.

orneryd:
Thanks for the thorough report and implementation. The complete-primary-key-group approach was the right foundation and fixes the original 720-row tied-key reproduction.

I reviewed the remaining edge cases and pushed a follow-up directly to the PR branch. It adds correctness guards for descending null ordering, cross-variable traversal sort keys, and unlabelled MATCH queries, plus stable index-group membership during concurrent mutation.

I also changed selection to use a group-local heap over only the secondary keys. Compared with the original PR head, the benchmark improved by approximately 49% in latency, 43% in allocated bytes, and 72% in allocations. Focused race tests and the complete Cypher/storage suites pass.

PR #375, opened 2026-09-14
feat(search): add secure cross-protocol qid continuation

orneryd:
# Durable Cross-Protocol QID Continuation Plan

## Objective

Extend NornicDB's Neo4j-compatible Bolt result-stream contract into one durable,
protocol-neutral continuation mechanism for ranked search.

The end state is:

- Bolt preserves standard `RUN`, `PULL {qid, n}`, and `DISCARD {qid}` behavior.
- A connection-local Bolt `qid` can alias the same retained result population
  used by HTTP, native gRPC, and Cypher.
- Extension-aware Bolt clients can obtain an opaque durable qid and resume the
  result after reconnecting or through another supported protocol.
- HTTP, native gRPC, and `db.retrieve` expose the same start, pull, and discard
  semantics instead of separate page and release APIs.
- Search continuation incrementally extends the final canonical outer-RRF
  population produced by `search.SearchTextChunks`; the initial retrieval depth
  is not a total-result ceiling.
- No database transaction, storage lock, search-index lock, or global registry
  lock remains held while the client decides whether to request another page.
- Search calls that do not request continuation retain their current behavior
  and performance.

This plan supersedes the API shape proposed by PR #353. Its `ranked`,
`ranked_then_id`, `id`, complete filtered scan, grouping, exhaustion, and
mutation-epoch semantics are carried forward through the shared qid contract.
Its separate `db.retrieve.page`/`db.retrieve.release` procedures, cursor store,
global locking, and deep-copy-heavy population representation are not.

## Contract

All protocols implement the same logical operations:

```text
START(query, options, n) -> records, qid, has_more
PULL(qid, n)             -> records, qid, has_more
DISCARD(qid)             -> released
```

Properties of this contract:

- `n` is only the maximum number of records returned by one operation.
- In continuation mode, `limit` is the initial retrieval depth and expansion
  quantum, not the maximum number of results the caller may consume.
- Optional `max_results` lets a caller request an explicit total ceiling. When
  omitted, pulls continue until the searchable population is exhausted or a
  configured server safety limit rejects further expansion.
- A successful partial response returns a qid representing the next position.
- Repeating the same durable qid is idempotent and returns the same page.
- `DISCARD` releases the complete retained population, regardless of which
  position token from that population is supplied.
- An exhausted response has `has_more: false` and no next qid.
- Pulling does not extend the fixed expiry time.
- Later pulls never rerun chunking or embedding. They may deepen ANN and BM25
[...]

PR #377, opened 2026-09-14
fix(search): preserve exhaustion and metadata in continuation

mm0nst3r:
Follow-up to #375, targeting `search-continuation`.

**The exhaustion fix applies to BM25, vector and hybrid search, including Voyage embeddings. End-to-end verification of the missing-results regression used BM25 only. Separate actual-Voyage generation/retrieval checks passed on the four-PR source; real-provider pagination has not been tested.**

A short filtered or approximate batch can end ranked continuation while deeper retrieval still finds matches. This change carries explicit exhaustion evidence from retrieval through fusion and continuation. BM25 uses its unfiltered exact prefix; exact vector paths use candidate coverage; HNSW requires its pre-filter heap to cover live index membership without top-k truncation. Continued multi-chunk retrieval can deepen beyond the ordinary prefix and reuse prepared embeddings.

Candidate budgets remain in place. Unknown coverage at a budget ceiling returns a capacity error instead of false completion. `max_results` limits emitted members without prematurely limiting retrieval depth. The fixed ranked prefix followed by catalogue enumeration in `ranked_then_id` is preserved.

Continuation also preserves the canonical search response and result metadata instead of reconstructing narrower copies. Grouped children use the same result model as top-level hits in Go and native gRPC; adapters retain metadata through provider-neutral accessors. This PR contains no Voyage client, embedding worker or reranker implementation. Voyage's metadata/coverage producer belongs to #355. Every substantive fix is in these PRs; no additional integration implementation is required after merging them.

Verification:
- Original BM25 and HTTP missing-results regressions fail on #375 and pass with this correction; focused tests cover approximate/filtered batches, HNSW deletion and fallback, chunk deepening, replay and budget errors.
- Current four-PR Linux CPU build and selected transport/search suites pass. Provider score filtering and grouped/paginated metadata are exercised with a simulated provider.
- HTTP and Bolt reuse 101 original Library records with both BM25 engines, stemming transitions, updates/deletion/recreation, exact text preservation and full 15-page ID/ranked-then-ID enumeration with replay.
- Actual Voyage transcript/JPEG generation, retrieval, lifecycle/restart, 101-candidate reranking and gRPC model override pass separately on the same combined source.

No production deployment or blanket whole-repository suite pass is claimed. The original nanosecond-timeout test is intermittent on Windows; it also fails on the previous #377 head. The configured native rerank pool still bounds progressive ranking and can produce an explicit capacity error when exceeded.


orneryd:
I implemented the fixes in the other branch there were more areas there. I'll take a look at the other issue

Issue #487, opened 2026-09-19
[PERF] Reads inside an explicit transaction are 2–11x slower than the same auto-commit statement: the transactional label scan materialises every node of the label (per-node MVCC head lookup + copy) before predicates, traversal or LIMIT


mm0nst3r:
## Performance Issue Description
The same read statement is 2–11× slower, and allocates 3–6× more, when it runs inside an explicit transaction (`BEGIN … COMMIT`, i.e. what `driver.execute_query` / `session.execute_read` use) than as an auto-commit statement. This is separate from #436 (that one was a scan of the *whole database*; this is after that fix, on main at 4ec45a85).

The cause visible in the profile: inside a transaction a label-scoped `MATCH` goes through `transactionStorageWrapper.GetNodesByLabel` → `BadgerTransaction.GetNodesByLabel` → `getNodesByLabelVisibleAtSnapshotWithView`, which resolves **every node of the label** one by one (`getNodeVisibleAtInTxn`: a point lookup of the MVCC head per node, version resolution, `copyNodeWithoutEmbeddings`) and returns a fully materialised `[]*Node` before any predicate, traversal or `LIMIT` is applied. The auto-commit path streams and can stop early. So the transactional read path does not follow the streaming / project-only-what-is-needed model the rest of the storage API is moving to.

## Environment
- **NornicDB Version**: main at 4ec45a85
- **OS**: `golang:1.27.1-bookworm` container, 4 CPUs, Docker Desktop / WSL2
- **Architecture**: AMD64 (Ryzen 9 7950X)
- **Build tags**: `noui,nolocalllm`

## Dataset Size
- In-memory engine behind `NamespacedEngine`: 2,000 `:Person` nodes (3 small properties, no embeddings), 500 `:KNOWS` relationships. No property index on `:Person(id)`.

## Query
```cypher
MATCH (n:Person) RETURN n.name LIMIT 10
MATCH (n:Person {id: 42}) RETURN n.name
MATCH (n:Person) WHERE n.age > 30 RETURN count(n)
MATCH (a:Person)-[:KNOWS]->(b:Person) RETURN a.name, b.name LIMIT 20
RETURN 1
```

## Performance Metrics
`go test ./pkg/cypher -run '^$' -bench BenchmarkStatementRouting -benchmem -benchtime=1s -count=5`, medians. An unused `$nonce` parameter changes on every iteration so the result cache never answers.

[...]

Issue #475, opened 2026-09-19
[BUG] Subscript on a list property returns null: `a.tags[0]` → null, so `WHERE a.tags[0] = 't0'` never matches


mm0nst3r:
## Bug Description
Indexing a list-valued node property yields null in `RETURN` and therefore never matches in `WHERE`. `size(a.tags)` and `'x' IN a.tags` work; `head(a.tags)` works in auto-commit but returns no match inside an explicit transaction.

## Steps to Reproduce
Fresh database; setup (three statements, each in its own transaction):
```cypher
CREATE (:A {id:1, name:'n1', tags:['t1','all']}), (:A {id:2, name:'n2', tags:['t0','all']}), (:A {id:3, name:'n3', tags:['t1','all']}), (:A {id:4, name:'n4', tags:['t0','all']})
MATCH (x:A {id:1}), (y:A {id:2}) CREATE (x)-[:NEXT {w:10}]->(y)
MATCH (x:A {id:2}), (y:A {id:3}) CREATE (x)-[:NEXT {w:20}]->(y)
```
```cypher
MATCH (a:A) RETURN a.id AS id, a.tags[0] AS first ORDER BY id LIMIT 2
MATCH (a:A) WHERE a.tags[0] = 't0' RETURN a.id AS id ORDER BY id
MATCH (a:A) WHERE head(a.tags) = 't0' RETURN count(a) AS c
```

## Expected Behavior
```
1) [{'id': 1, 'first': 't1'}, {'id': 2, 'first': 't0'}]
2) [{'id': 2}, {'id': 4}]
3) [{'c': 2}]
```

## Actual Behavior
```
1) [{'id': 1, 'first': None}, {'id': 2, 'first': None}]
2) []
3) auto-commit:  [{'c': 2}]   (correct)
              explicit tx:  [{'c': 0}]
```
No error is returned in any of these cases.

## Environment
- **NornicDB Version**: main at 994b3a68
- **OS**: Linux container (debian:bookworm-slim) on Docker Desktop / WSL2
- **Architecture**: AMD64
- **Docker Image**: built from source, `-tags "noui,nolocalllm"`, embeddings disabled
- **Driver**: neo4j Python driver 5.28, Bolt

## Additional Context
Subscripting a literal list works (`[1,2,3][1]` → 2).

## Checklist
- [x] I have searched existing issues to ensure this is not a duplicate
- [x] I have tested this with the latest version of NornicDB
[...]

Issue #461, opened 2026-09-19
[BUG] A relationship created together with its nodes in one auto-commit CREATE is invisible to pattern matches that bind it inside explicit transactions (and to MATCH … SET), permanently


mm0nst3r:
## Bug Description
A relationship that is created **together with its two nodes in one Bolt auto-commit statement** (`CREATE (a)-[:R {…}]->(b)`) is permanently invisible to pattern matches that bind the relationship inside an explicit transaction (`BEGIN`/`RUN`/`COMMIT`, i.e. `driver.execute_query`, `session.execute_read/execute_write`), and to auto-commit statements that write (`MATCH ()-[h]->() SET …`, which run in an implicit transaction). Auto-commit reads see it. The nodes are visible in both modes, and patterns that do not bind the relationship (`(o)-->(i)`, `count(h)` on an anonymous pattern) see it too.

If the two nodes are created first and the relationship in a second statement, or if everything is created inside an explicit transaction, the relationship is visible everywhere. It is not a timing issue: the result is the same 5 s later. It is not a regression of the recent transactional-read work either: 8672a400 behaves the same.

Consequence: data loaded with plain `session.run("CREATE (a)-[:R]->(b)")` looks fine in ad-hoc reads, but application code using managed transactions gets no rows, and relationship updates silently match nothing (`matched: 0`).

## Steps to Reproduce
Fresh database, embeddings disabled, neo4j Python driver 5.28:
```python
# A relationship written by one Bolt auto-commit statement, read back from an explicit transaction.
import sys, time
from neo4j import GraphDatabase
d = GraphDatabase.driver(f'bolt://{sys.argv[1]}:7687', auth=None)
def auto(q, **p):
    with d.session() as s: return [r.data() for r in s.run(q, **p)]
def tx(q, **p): return [r.data() for r in d.execute_query(q, **p).records]     # BEGIN / RUN / COMMIT
def case(title, writes):
    auto('MATCH (n) DETACH DELETE n')
    for w in writes: auto(w)
    time.sleep(1)
    print(title)
[...]

PR #488, opened 2026-09-19
refactor(cypher): route statements inside explicit transactions through the main router

mm0nst3r:
## What
`executeQueryAgainstStorage` (the statement dispatcher used only inside `BEGIN … COMMIT`) was a second, separately maintained copy of the keyword dispatch in `executeWithoutTransaction`, with a different rule order and without several handlers and fast paths. This PR removes that copy: inside an explicit transaction the statement now goes through the same router as auto-commit, and the transaction only supplies the storage view. The reveal/decay preamble is kept.

The main router already runs against the transactional storage wrapper for every auto-commit write (`executeWithImplicitTransaction` → `txExec.executeWithoutTransaction`), so this is not a new combination.

`pkg/cypher/transaction.go`: +2 / −204.

## Behaviour changes (as agreed: auto-commit behaviour is the reference)
Inside an explicit transaction:
- unknown `SHOW …` now returns the main router's error (`unsupported query type: SHOW …`) instead of the transaction router's own wording;
- `ALTER COMPOSITE DATABASE … ADD/DROP CONSTITUENT` is supported (`executeAlterCompositeDatabase`), but the transaction router did not know it and refused it as `unsupported query type`; it now reaches the real handler inside a transaction as well. (The updated test expects `requires multi-database support` only because that unit test's executor has no database manager.)
- `LOAD CSV` returns the main router's message (`LOAD CSV is not supported in NornicDB embedded mode`) instead of `unsupported query type`;
- `DROP INDEX <missing>` is an error (`does not exist`) and `DROP INDEX <missing> IF EXISTS` succeeds, as in auto-commit; before it was a silent no-op;
- shapes that only the main router handled (`MATCH … CALL {}` subqueries, multiple `MERGE`, `ALTER DATABASE`, the simple-match and compound fast paths) are now reachable inside a transaction.

Three expectations in `executor_misc_test.go` and two cases in `transaction_namespace_test.go` pinned the removed router's wording / no-op and were updated to the above. No test was removed.

Added `transaction_bare_begin_test.go` (27 cases): after a bare `BEGIN`, and in the one-request `BEGIN … COMMIT|ROLLBACK` script form, each write shape that takes a different route in auto-commit (pure `CREATE`, `CREATE` with a relationship, `MERGE`, `MATCH … SET`, `MATCH … CREATE`, `UNWIND … CREATE`) runs in the transaction `BEGIN` opened: same transaction object and id before and after the statement, own writes visible inside, nothing visible to a second session before `COMMIT`, nothing left after `ROLLBACK`. They pass on this branch and on unmodified main.

## Tests
- `go build ./...`, `go vet ./pkg/cypher`: clean.
- `go test ./pkg/cypher/... ./pkg/bolt/... ./pkg/server/... ./pkg/nornicdb/...`: pass.
- 131 Cypher shapes executed over Bolt in auto-commit and in explicit-transaction mode against a server built from this branch: result sets identical to main in both modes.

## Performance
New `BenchmarkStatementRouting` (`pkg/cypher/transaction_router_bench_test.go`): five statements, auto-commit and explicit transaction, 2,000 nodes / 500 relationships, result cache defeated with a changing parameter. Median of 5, main → this branch:

| statement | auto-commit | explicit tx |
| --- | --- | --- |
| simple_match_limit | 25.7 → 25.3 µs | 52.6 → 56.3 µs |
| property_lookup | 1146 → 1143 µs | 5827 → 6073 µs |
| filter_count | 1433 → 1419 µs | 5898 → 5813 µs |
| one_hop_limit | 977 → 1011 µs | 10554 → 10946 µs |
| return_literal | 3.8 → 3.6 µs | 9.5 → 10.0 µs |

The per-run ranges of main and branch overlap in every row (e.g. simple_match_limit 51.7–55.0 vs 53.8–61.5 µs; one_hop 10.4–11.1 vs 9.4–12.6 ms), and allocations are unchanged in auto-commit and +6 to +21 allocations per statement inside a transaction (of 47 to 88,610), so this is neutral within noise. The cost of a statement inside a transaction is in the storage view, not in routing; that is tracked separately in #487.

## Not in this PR
The remaining differences between the two modes (#461, #475) come from the storage view, not from routing, and are unchanged here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mm0nst3r:
Added the bare-`BEGIN` test (`pkg/cypher/transaction_bare_begin_test.go`, 19afc38c).

What it pins: a statement that follows `BEGIN` runs in the transaction `BEGIN` opened; it never opens, commits or replaces a transaction of its own. Checked for six write shapes that each take a different route in auto-commit (pure `CREATE`, `CREATE` with a relationship, `MERGE`, `MATCH … SET`, `MATCH … CREATE`, `UNWIND … CREATE`), each ended with `ROLLBACK` and with `COMMIT`, in two forms: separate statements after a bare `BEGIN`, and the one-request `BEGIN <query> COMMIT|ROLLBACK` script.

Assertions per case: same transaction object and txID before and after the statement and still active; the session sees its own write; a second session on the same store sees nothing before `COMMIT`; after `ROLLBACK` nothing is left (a nested commit would leave data behind); after `COMMIT` the second session sees exactly what the first saw. Plus one case where a statement fails after an earlier write and the client rolls back.

27 cases, all pass on this branch and on unmodified main, so this is a guard for the behaviour, not a behaviour change.

On `ALTER COMPOSITE DATABASE`: it is supported (`ADD CONSTITUENT` / `DROP CONSTITUENT`, `executeAlterCompositeDatabase` in `composite_commands.go`), not only create and drop. The removed transaction router did not have a rule for it and refused it as `unsupported query type`; with one router it reaches the real handler inside a transaction too. The `requires multi-database support` expectation in the updated unit test is only because that test's executor is built without a database manager. I corrected the PR description, which was misleading on this point.

orneryd (approved):

